PR #59, opened 2015-02-12
added test for sqlite using the file system

ilvalle:
The current sqlite test uses sqlite:memory by doing that we are not able to run test against table migration for sqlite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
     - PIP_DOWNLOAD_CACHE=$HOME/.pip-cache
   matrix:
     - ADAPTER=sqlite
+    - ADAPTER=sqliteFS
     - ADAPTER=mysql
     - ADAPTER=postgres
     - ADAPTER=postgresPG8000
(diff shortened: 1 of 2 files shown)


gi0baro:
Maybe we can replace the sqlite:memory with the fs test? Thinking on when I will push the py3 branch and tests, and I think we can re-consider the build matrix on travis and avoid the repetitive ones, as pointed @niphlod in another PR comments..

mdipierro:
ok

niphlod:
yep, my point is that now matrix is quite large. It's true that we're leveraging the docker infrastructure on travis that is super-fast, but maybe at this point we should consider running different test runs in an env that doesn't stand for "module imported" but with "services that can be used". 

BTW: unrelated... a few builds have failed because tests run so fast that coveralls isn't ready to accept the coverage reports. Some times it's not the tests that are failing, rather the upload on coveralls. We should consider to move the coveralls command to an after_success script rather than having travis directly executing it.

gi0baro:
@niphlod +1 for the `after_success` script. Can you make a PR about that?
In the meantime I will spend some time on thinking about the whole matrix thing.

niphlod:
No problem. Wait a few minutes.

niphlod:
there it is https://github.com/web2py/pydal/pull/77

PR #71, opened 2015-03-07
fix common filter in mongo adapter

ilvalle:
reported here https://groups.google.com/d/topic/web2py-developers/H3sk7RqCtYQ/discussion

### DIFF
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -6,7 +6,7 @@ import re
 from .._globals import IDENTITY
 from ..objects import Table, Query, Field, Expression
 from ..helpers.classes import SQLALL
-from ..helpers.methods import xorify
+from ..helpers.methods import use_common_filters, xorify
 from .base import NoSQLAdapter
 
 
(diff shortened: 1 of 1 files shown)


mdipierro:
Travis failed but I do not see any change in this code that would cause that fail. I think it a travis error. Unless there is an objection I am going to merge this.

niphlod:
you can restart any travis build if you feel it's an error, like it was on this case.

BTW: I feel that a matrix composed by 23 builds is kinda unneeded for pydal.

ilvalle:
Regarding running tests, web2py has several imports in a try-catch statement and several contrib modules, as a matter of fact what is really missing in the current matrix are tests against all possible imported modules. 
Given the following:
```
try:
    import module_a
except:
    from .contrib import module_a
```
In this case, we should run two tests, the one using module_a and then one using the contrib version.
By doing that, we can guarantee that crontib modules shipped with web2py are updated and inline with the pip version. As as happened recently for portalocker, they are not always inline.
I don't want to run an infinitive combination of the possible modules, just a few for the most important ones.

franciscocosta:
this fix solves the problem.
But I found another issue (in fact is a old one): https://groups.google.com/d/msg/web2py/0tVp5QVR1-8/8F4Xojdug7QJ

Is it possible to have MongoDB with versioning and ondelete?

ilvalle:
@franciscocosta open an other issue for the new problem

PR #82, opened 2015-03-13
Better adapters and gae imports flow

gi0baro:
As discussed on 71568d8

### DIFF
--- a/pydal/_gae.py
+++ b/pydal/_gae.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import decimal
 
 try:
     from new import classobj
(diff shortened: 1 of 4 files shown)


pydal/_gae.py
```diff
@@ -83,3 +78,6 @@ def _validate(self, value):
                 return decimal.Decimal(value)
             raise TypeError("Property %s must be a Decimal or string."
                             % self._name)
```
mdipierro: I am ok with the other changes but I think this file should stay as it is.

gi0baro: Any technical reason to do that?
With the changes on *adapters/\__init\__.py*, if something is wrong with the google adapters you will se the errors if your concerns are about this.

mdipierro: Mi concern is to have a lot of code into try ... except just to detect if we have GAE.

gi0baro: And why is that a problem? If the first line of the `try` block fails the interpreter will stop the execution and jump to the `except` one, no matter of how many lines you have inside the block..

mdipierro: The problem is that if the first line succeeds (we have GAE) and one of the
latter fails because we have a bug in those lines, or a missing module,
etc, it will think we are not on GAE and we do not get a traceback

mdipierro:
I merged part of this manually.

pydal/_gae.py
gi0baro: I see your point. let me update this PR with a new solution.

gi0baro:
I'm going to undo the partial and update the PR

gi0baro:
@mdipierro what do you think of this solution?

PR #89, opened 2015-03-18
common_filter fix for delete/update on mongo adapter and tests

ilvalle:
The build fails due to the new pg8000 driver released the 17th of March.


ilvalle:
I created the issue for pg8000 https://github.com/mfenniak/pg8000/issues/76

Issue #112, opened 2015-03-27
datetime not json serializable


salomonderossi:
Datetime objects in the pyDLA are not json serializable. In a pevious version of web2py's DAL json serializion did work fine.

I use Version 15.3 of pyDAL with python2.7 on Ubuntu 14.04.2 LTS.

Here ist a small program which demonstrates the issue:
```python
from pydal import DAL, Field
from datetime import datetime

db = DAL('sqlite://storage.db', folder="test_databases")
db.define_table('test_table', Field('date_field', 'datetime'))

db.test_table.insert(date_field=datetime.now())

rows = db().select(db.test_table.ALL)
print(rows.as_json())
```
Which gives this output on my machine:
```
Traceback (most recent call last):
  File "db_text.py", line 13, in <module>
    print(rows.as_json())
  File "/usr/local/lib/python2.7/dist-packages/pydal/objects.py", line 2741, in as_json
    return json.dumps(items)
  File "/usr/lib/python2.7/json/__init__.py", line 243, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: datetime.datetime(2015, 3, 27, 13, 37, 7) is not JSON serializable
```

PR #113, opened 2015-03-27
test and fix as_json when web2py serializer is not available, closes web2py/pydal#112

ilvalle:
I'd consider to backport https://github.com/web2py/web2py/blob/master/gluon/serializers.py (or part of it) in pydal.

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -2726,13 +2726,13 @@ class Rows(object):
         mode='object' is not implemented (should return a nested
         object structure)
         """
-
+        has_serializer = self.db.has_serializer('json')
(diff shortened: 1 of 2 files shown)


gi0baro:
I'd like to discuss this a bit before a merge, since as you reported we should decide if and how to back port part of the serialization we have in web2py.

mdipierro:
Oops. I merged too soon. Let's discuss and then revert if needed. What do you recommend?

ilvalle:
The alternative is to backport custom_json(), json(), load_json() from web2py

ilvalle:
The alternative is to backport custom_json(), json(), load_json() from web2py

PR #123, opened 2015-04-03
Updated BaseAdapter.represent to resolve web2py issues/697

ProjectGuerilla:
This is a patch for the error caused by putting unicode strings into a Field of type list:string. https://github.com/web2py/web2py/issues/697

### DIFF
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -1342,7 +1342,7 @@ class BaseAdapter(ConnectionPool):
             elif not isinstance(obj, (list, tuple)):
                 obj = [obj]
             if field_is_type('list:string'):
-                obj = map(str,obj)
+                obj = map(lambda x:unicode(x).encode(self.db_codec),obj)
             else:
                 obj = map(int,[o for o in obj if o != ''])
         # we don't want to bar_encode json objects


gi0baro:
Thank you for the solution, but I have a couple of considerations:
- This patch won't work on python3, and we're working on it.
- I'm not a big fan of always cast the objects to `unicode` in order to solve a bad usage

I think we should work on a better implementation for this.

ProjectGuerilla:
Those are great comments - I'm not sure how to avoid casting to unicode since the original code casts to string ;-). Since I could not immediately see WHY it is being cast to string I didn't know where to begin in looking for a more systemic change.

Clearly there are some deeper changes that will be needed to avoid this problem. If I have time to dig deeper into the code I will see what I can uncover.

ProjectGuerilla:
OK, I have look a *little* more closely and the bar_encode call requires a *string* AND it uses the older, python 2 style string formatting
```
'|%s|' % '|'.join(bar_escape(item) for item in items if str(item).strip())
```
so the comment about not working in Python3 seems somewhat ahead of the game right now. If there is a complete review required in order to move to Python3, then the unicode call will get removed then (the code could even be commented to give the Python3 code required).

As for NOT wanting to cast to unicode - we are getting arbitrary text and converting it into a string with some flavor or unicode encoding that will handle the passed values. Eventually we need to have the arbitrary text converted to a string encoding the is of the same flavor as the database. I suppose it could be wrapped in a try/except to save us from an arbitrary conversion to unicode but the database has an encoding (UTF-8 by default).

Would it be better to check the database encoding and only convert to unicode if the database is NOT ascii?
```
try:
    obj = map(str,obj)
except UnicodeEncodeError:
    obj = map(lambda x:unicode(x, self.db_codec).encode(self.db_codec),obj)
```

ProjectGuerilla:
Also, please DON'T pull this request! there is an error I found in the code…

**IF** I get some positive feedback I'll make the patch and resubmit.
'''
obj = map(lambda x:unicode(x).encode(self.db_codec),obj)
'''
should be:
'''
obj = map(lambda x:unicode(x,self.db_codec).encode(self.db_codec),obj)
'''

mdipierro:
Please do

ProjectGuerilla:
I have made a change that improves on this pull request slightly - please consider 
https://github.com/web2py/pydal/pull/131
instead of this pull request.

PR #125, opened 2015-04-03
fix drivers.py for ironpython support

sp1d3rx:
Ironpython throws an 'attributeerror' instead of an importerror on importing a function from a library. I adjusted this so that it will catch both instead of erroring out.

### DIFF
--- a/pydal/drivers.py
+++ b/pydal/drivers.py
@@ -14,50 +14,50 @@ else:
     try:
         from pysqlite2 import dbapi2 as sqlite2
         DRIVERS['sqlite2'] = sqlite2
-    except ImportError:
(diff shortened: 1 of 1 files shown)


mdipierro:
I think IronPython is not following the specs here. This is a bug in IronPython and should be addressed there, not here. At this time the core developers are not very keen in supporting IronPython. Please make a case for why we should support IronPython on the web2py developers mailing list.

sp1d3rx:
I'll leave it as it is, take it or leave it. I just wanted to do what I can to provide support for IronPython. If you think that capturing the AttributeError on import is too much of a hassle to handle, then that's fine by me.

PR #133, opened 2015-04-09
fix mongo adapter compatibility of with pymongo 3.0

ilvalle:
A first iteration that passes all tests using new pymongo. There is still room for improvements (i.e. to handle better the safe and snapshot parameters).

- deprecated methods have been replaced with the new ones (as insert_one, update_many etc)
- count is now implemented using pymongo count()


gi0baro:
@ilvalle did you tested backward compatibility with pymongo 2.x?

ilvalle:
It isn't compatible. Mongo Adapter it's consider as experimental, which allows us to update it without considering backward compatibility (luckily I'd say). We should force users to use the latest.

PR #139, opened 2015-04-14
added Rows.__eq__ and few tests

ilvalle:


pydal/objects.py
```diff
@@ -2386,6 +2386,12 @@ def __str__(self):
         self.export_to_csv_file(s)
         return s.getvalue()
 
+    def __eq__(self, other):
+        if len(self.records) != len(other.records):
+            return False
+        l = map(lambda x: x[0] == x[1], zip(self.records, other.records))
```
niphlod: given the recent "discoveries" of iterRows and the fact that DAL can be less memory hungry if coded efficiently....you took the time to code an __eq__: you should just break at the first difference found, without looping through the entire resultset and then checking if all the checks are True or False...think 1M:1M Rows to compare......

niphlod: benched a few methods (yours, izip, izip + imap, all(imap(izip)), etc). memory utilization is the same once you pull in generators.......seems that sticking to izip is the fastest, and is the closes to the native "list == list" check...now, assuming that self.records == other.records is not feasible (wondering why)... 
I tested with 1M records, plus a different record placed at the very end. 
map() is ~25x slower than list==list, and 3x memory hungry
izip is ~9x slower, 1x memory
all(imap(izip)) is ~16x slower , 1x memory
So, here the "izip" version, that is quite readable, too
```
from itertools import izip
....
    if len(self.records) != len(other.records):
        return False
    for x in izip(self.records, other.records):
        if x[0] != x[1]:
            return False
    return True
```

ilvalle:
Thanks for the comment, I switched to the native list==list

PR #142, opened 2015-04-20
Added support for cache dictionary for select method. This allows finer-...

rpwatkins:
Added support for cache dictionary for select method. This allows finer-grained cache invalidation. This matches the changes to the count method in objects.py.

### DIFF
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -1207,9 +1207,17 @@ class BaseAdapter(ConnectionPool):
             self.execute(sql)
             rows = self._fetchall()
         else:
-            (cache_model, time_expire) = cache
-            key = self.uri + '/' + sql + '/rows'
-            key = hashlib_md5(key).hexdigest()
+            if isinstance(cache, dict):
+                cache_model = cache['model']
+                time_expire = cache['expiration']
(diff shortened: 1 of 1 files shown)


niphlod:
add tests if you want to change / modify / add something to the existing API.

pydal/adapters/base.py
```diff
@@ -1207,9 +1207,17 @@ def _select_aux(self, sql, fields, attributes):
...
         else:
-            (cache_model, time_expire) = cache
-            key = self.uri + '/' + sql + '/rows'
-            key = hashlib_md5(key).hexdigest()
+            if isinstance(cache, dict):
+                cache_model = cache['model']
+                time_expire = cache['expiration']
+                key = cache.get('key')
+                if not key:
+                    key = self.uri + '/' + sql
```
ilvalle: Mind that this concatenation is different from the one we have few lines below

ilvalle:
At this point, we should refactor the code in order to delegate the creation of the key to a single method

mdipierro:
I will take the patch, make the recommended changes but we should eventually follow ilvalle's advice.

gi0baro:
what about the tests, as mentioned by @niphlod ?

PR #147, opened 2015-04-25
dropped contrib/simplejson

ilvalle:


pydal/_load.py
```diff
@@ -5,10 +5,12 @@
     from collections import OrderedDict
 except:
     from .contrib.ordereddict import OrderedDict
+    
 try:
+# try external module
+    import simplejson as json
```
gi0baro: why keep trying to load simplejson if we're dropping py25?

ilvalle: as far as I remember, the external module (if present) is faster

 Paolo

2015-04-27 0:04 GMT+02:00 Giovanni Barillari <notifications@github.com>:

gi0baro: From simplejson page: 
" simplejson is the externally maintained development version of the json library included with Python 2.6 and Python 3.0, but maintains backwards compatibility with Python 2.5."

Btw, this is already patched in py3 branch.

niphlod: simplejson is updated more frequently and is generally faster than the json module shipped with python. It's not going to hurt anybody if you leave that line, importing simplejson if it's found and fallbacking to json.

gi0baro:
Since this is already done in py3 and I'm gonna merge it, I'm closing this

PR #148, opened 2015-04-27
added BasicStorage

ilvalle:
BasicStroage has the performance of web2py FastStorage without the memory leak.
DAL, Row & Table inherit from BasicStorage

Fix https://github.com/web2py/web2py/issues/932

### DIFF
--- a/pydal/base.py
+++ b/pydal/base.py
@@ -123,19 +123,24 @@ For more info::
 
 """
 
-import threading
-import socket
-import urllib
-import time
 import copy
-import traceback
 import glob
 import logging
+import socket
+import threading
+import time
+import traceback
+import urllib
 from uuid import uuid4
+
 
 from ._compat import pickle, hashlib_md5, pjoin, ogetattr, osetattr, copyreg
 from ._globals import GLOBAL_LOCKER, THREAD_LOCAL, DEFAULT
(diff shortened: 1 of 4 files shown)


gi0baro:
Why *BasicStorage2* and not *BasicStorage*? Probably I would named it *DALobject*..

ilvalle:
BasicStorage is more appropriate.
In my view, by changing only a couple of methods it can easily became the
super class of web2py Storage, this is the reason of naming it as
BasicStorage. The difference is that, BasicStorage raise an
attribute/keyError when an attribute/item is not available, Storage return
None. Anyway, this is not fully related to this PR

 Paolo

2015-04-27 8:13 GMT+02:00 Giovanni Barillari <notifications@github.com>:

gi0baro:
Ok, but you named the class *BasicStorage2*..

ilvalle:
I can't rename it right now. Beside the name, if you accept the patch,
merge it. I'll rename the class later today or tomorrow.

 Paolo

2015-04-27 9:01 GMT+02:00 Giovanni Barillari <notifications@github.com>:

gi0baro:
Ok, I'm gonna accept this, and rename the class in py3 branch with python3 fixtures

PR #151, opened 2015-04-27
fix delete_uploaded_files for computed fields

ilvalle:
A common thumbnail is defined as follows, The file is not deleted after update/delete. This PR fix it.
```
    Field("thumbnail", "upload", uploadfolder=os.path.join(request.folder,'uploads'),
          autodelete=True,  requires=[IS_IMAGE()],
          compute = lambda row: THUMBER(row.picture, 250, 200, name='thumb'))
```


gi0baro:
@mdipierro You merged this, but py3 is broken.

PR #158, opened 2015-04-28
Added BasicRows, fix py3 StringIO issue

ilvalle:
Added as_dict() as_csv() as_json() as_xml() as_list() to IterRows

### DIFF
--- a/pydal/_compat.py
+++ b/pydal/_compat.py
@@ -8,7 +8,7 @@ _identity = lambda x: x
 
 if PY2:
     import cPickle as pickle
-    import cStringIO as StringIO
+    from cStringIO import StringIO
     import copy_reg as copyreg
     hashlib_md5 = hashlib.md5
     iterkeys = lambda d: d.iterkeys()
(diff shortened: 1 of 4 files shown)


gi0baro:
I would ask why you need `abc` and `ABCMeta` for this?

Beside the StringIO fix, which I like,  there are quite a lot of stuffs you wrote that are not python3 compatible, please check out https://github.com/web2py/pydal/wiki/Write-py2py3-compatible-code

ilvalle:
Actually, I've just refactored Rows by introducing the super class BasicRows, and I added a few tests, many methods weren't tested at all (this is the reason because I spotted the StringIO issue)
I've introduced abc to explicitly mark the class as Abstract, now I removed it since it isn't compatible with py3

gi0baro:
Actually `abc` is python3 compatible, the objection was that is pointless use it without implements the decorators inside the class. Anyway, I will merge this and then update the code with python3 corrections.

PR #160, opened 2015-04-29
fix mssql with py3

niphlod:
let's see if this suffice. Please wait before actually merging.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,8 +24,13 @@ install:
   - pip install codecov
 
 before_test:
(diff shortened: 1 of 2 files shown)


gi0baro:
Seems not.. :(

niphlod:
no worries, it's a weird day for ci servers. Seems that MSSQL is not ready. Going to fix appveyor.yml to wait for it. let's see how it goes.

niphlod:
the syntax error gone away but for some reason py3 tests block, and the build errors out for timeout rather than an error. Please merge this in and I'll investigate further as soon as I get hands on py3.

PR #162, opened 2015-04-29
Unittest/smart query

BuhtigithuB:
There is some commented test case which are worth to be analysed to make sure there is no "bug" in smart_query. I comment stuff for now to make this PR so we can have the other test and raise a bit test coverage. 

:)

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -7,3 +7,4 @@ else:
 
 from .validation import *
 from .caching import TestCache
+from .test_smart_query import *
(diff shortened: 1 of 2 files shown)


gi0baro:
@ilvalle @niphlod Seems this is not working on noSQL adapters. Can you review this?

ilvalle:
Hi, on mongodb the first failing assertion is due to:
```<Query {'boolean_field': 'true'}> != <Query {'boolean_field': True}>```

I'll look at the issue on nosql in the following days, in the meanwhile wrap the whole test with @unittest.skipIf(NOSQL, "Skip nosql")

BuhtigithuB:
Paolo, you plan to do it once the PR is merged or better I close this PR do
the change and make another PR?

Richard

PR #166, opened 2015-04-30
reorder operators tuples in smart_query

BuhtigithuB:
I reorder the operators so they are the same in test_smart_query and I find it more easy to make sure they are all they there (and there alias) with the way I grouped them


BuhtigithuB:
operators can't be reorder :(
Since it has to iterate over the tuples list and if "equal" is check before "equal to" it will be turn into "= to"

PR #180, opened 2015-05-05
Improve test coverage for MongoDB

stephenrauch:
Remove test case skipping for most existing test cases for Mongo and add
a few more from the Sql test cases.  Fix mongo driver for said test
cases.

### DIFF
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -166,7 +166,7 @@ class MongoDBAdapter(NoSQLAdapter):
                 return value
             from bson import Binary
             if not isinstance(value, Binary):
-                return Binary(str.encode(value,'utf-8'))
+                return Binary(value)
             return value
         elif isinstance(fieldtype, basestring):
             if fieldtype.startswith('list:'):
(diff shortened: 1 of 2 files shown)


stephenrauch:
This pull request failed the tests for GAE and python 3.  Is there a way I can invoke the tests?

gi0baro:
You should inspect travis logs, https://travis-ci.org/web2py/pydal/jobs/61231974 for GAE and https://travis-ci.org/web2py/pydal/jobs/61231987, to run those tests locally you should need mongo db running, and I think GAE is not testable locally.

For what I see, you simply added some tests that should be skipped on GAE, just add the forgot decorator. For the python3 issue on mongo, I think you changed the `represent` code to work with python2 (you used some `str()` calls), but the resulting objects in python3 are not bytes..

stephenrauch:
OK, figured out how to get the Travis CI to run, and sorted the GAE and python 3 problems.  The python 3 problems were actually from the test cases I enabled. Some failed under python 3, which I wasn't testing locally. The str() I added wasn't the problem, it was just the solution? for python 2 test cases.  The str() I added was the clue to the actual problem, and now with some reading I have a better understanding the python2/3 differences.

Issue #88, opened 2015-03-17
Duplicated signatures in mongo adapter


ilvalle:
The methods **LIKE**, **STARTSWITH**, **ENDSWITH**, **CONTAINS** are defined twice.
For each of them, which is the one to delete and the one to use?

niphlod:
I see that first occurrences handle more cases than the last. Dropping the last (and verifying that tests pass) should be the easy way to verify that,

ilvalle:
I'll verify because it should be the last occurrence the one used

ilvalle:
Actually, there are no tests that covers such methods https://coveralls.io/builds/2122836/source?filename=pydal%2Fadapters%2Fmongo.py

stephenrauch:
@ilvalle, did the recent work on **LIKE** and **CONTAINS** resolve this bug?

ilvalle:
yes of course.

PR #182, opened 2015-05-05
Fix contains for mongo db

stephenrauch:
For MongoDB Driver, restructure contains(), like(), ilike(), startswith() and endswith() to use consistent regex construction and to allow test cases to pass.  This sits on top of https://github.com/web2py/pydal/pull/180, who's changes were necessary for this change to pass its test cases.  This can make this change look way busier than it is.

### DIFF
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -544,70 +544,74 @@ class MongoDBAdapter(NoSQLAdapter):
                                   " but can be simulated with a wrapper.")
         return '%s ON %s' % (self.expand(first), self.expand(second))
 
-    # BLOW ARE TWO IMPLEMENTATIONS OF THE SAME FUNCITONS
-    # WHICH ONE IS BEST?
-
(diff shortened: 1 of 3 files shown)


ilvalle:
@stephenrauch well done. It should fix https://github.com/web2py/pydal/issues/88. 
You should rebase your branch against master in order to squash all your test commits into a few one.

stephenrauch:
OK, so I am new to GitHub and Git, and I just googled rebase and got the following:

> Warning: Because changing your commit history can make things difficult for everyone else
> using the repository, it's considered bad practice to rebase commits when you've already
> pushed to a repository.

I assume I can ignore this since this repository only has my changes.  Is this correct?

Also, should I do anything with https://github.com/web2py/pydal/pull/180 or just pull it all from here.

gi0baro:
@stephenrauch yes, you can rebase your branch without bothering.
If this PR integrates #180, you should close that.

PR #187, opened 2015-05-12
codecov now supports appveyor without a token

niphlod:
let's see.


niphlod:
can you please pull this is to see if codecov works with appveyor without explicitely set a token ? it can be reverted if not.

PR #191, opened 2015-05-14
requested by appveyor support

niphlod:
list available MSSQL driver DLL

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,7 @@ clone_depth: 50
 init:
   - "ECHO %PYTHON%"
   - set PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
+  - dir c:\windows\system32\sql*
 
 install:
   - ps: Start-FileDownload https://bootstrap.pypa.io/get-pip.py
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,3 @@
-os: Previous Windows Server 2012 R2
-
 build: false
 
 services:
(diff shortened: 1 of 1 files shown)


niphlod:
@gi0baro : can you wrap your head around it ? see https://github.com/appveyor/ci/issues/268#issuecomment-101778422

gi0baro:
@niphlod Windows is definitely not my "zone", but if it's not tox dependent, I will try to figure out what's happening

gi0baro:
@niphlod py27 is passing, I think was tox related. Will merge this if everything will pass ^^

niphlod:
pinpointed to tox >= 2.. 
Definitely.
``(new) introduce environment variable isolation: tox now only passes the PATH and PIP_INDEX_URL variable from the tox invocation environment to the test environment and on Windows also SYSTEMROOT, PATHEXT, TEMP and TMP whereas on unix additionally TMPDIR is passed. If you need to pass through further environment variables you can use the new passenv setting, a space-separated list of environment variable names. Each name can make use of fnmatch-style glob patterns. All environment variables which exist in the tox-invocation environment will be copied to the test environment.``

seems that both pyodbc and pypyodbc are relying on `WINDIR` to be defined. For all intents and purposes `passenv = WINDIR` added to tox.ini fixes the issue with tox >= 2, but I'm not entirely sure it's safe to do on Linux. For the time being please merge this in. Tomorrow I'm going to push a big piece fixing unicode in MSSQL.

PR #192, opened 2015-05-18
support unicode in MSSQL

niphlod:
plus CI integration fixes. As always, let's see if it's all fixed :D


niphlod:
btw: this is being used to debug codecov misbehaviours. Sorry for the long list of commits. I'll probably push again here and prepare a new rebased PR with all the fixes. Please don't merge this.

PR #193, opened 2015-05-18
Unicode support for MSSQL

niphlod:
plus a few fixes for CI envs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,8 +55,8 @@ before_script:
   - if [[ $ADAPTER == postgres* ]]; then psql -c 'SHOW SERVER_VERSION' -U postgres; fi
 
   # Install last sdk for app engine (update only whenever a new release is available)
-  - if [[ $ADAPTER == google ]]; then wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.19.zip -nv; fi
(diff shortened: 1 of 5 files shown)


niphlod:
ok, this is good to be merged. 
Notes: mssql2 was bugged in executesql, so it's fixed

notes^2 : all new adapters are experimental, but they pass tests on my rig (and on appveyor). There are no tests specifically for unicode, but in my trials everything went fine. 

in the following statements, deprecate means "don't support anymore/no new features/etc"

Technically there are too many adapters right now, and:
- `mssql2`: didn't touch anything that didn't break yesterday. In my vision it remains an experiment, it's there for backwards compatibility but should be deprecated ASAP
- `mssql` : default legacy for sql server 2000 --> has limitations because of "text" fields not being LIKE(able) --> contains() list:reference, list:string, etc methods are unavailable. Plus, it doesn't allow legacy unicode database handling. To be deprecated ASAP (if not right now)
- `mssqln` : all of the above except for handling unicode properly
- `mssql3, mssql4`: no "methods" limitations, but pagination improvements (4 better than 3). Again, they should have been marked experimental, but given that they don't handle unicode properly, they are there just to support non-unicode aware legacy databases, or web2py-managed ones created with those adapters.

`mssql3n` and `mssql4n` (after a bit of testing) should become the "defacto standard" for mssql databases.

gi0baro:
@niphlod Nice job!

Issue #178, opened 2015-05-04
Update mysql adapter `lastrowid()`


gi0baro:
As pointed out @niphlod on https://github.com/web2py/web2py/issues/667, we can use the better `cursor.lastrowid`.

For more infos: https://www.python.org/dev/peps/pep-0249/#lastrowid

gi0baro:
Note: this is only a *stylistic* choice, the resulting behavior should be the same.

ilvalle:
I agree on using ```lastrowid``` but double check that works on all major
drivers. My experience is that not all static attributes are available on
all drivers. For example ```rowcount``` doesn't work on all drivers.

 Paolo

2015-05-05 1:14 GMT+02:00 Giovanni Barillari <notifications@github.com>:

PR #194, opened 2015-05-20
introduction of cursor.lastrowid fixed web2py/pydal#178

ilvalle:


niphlod:
the added [test](https://github.com/web2py/pydal/pull/194/files#diff-3b2d9f3a86b72f767916790fb46cc94eR279) is quite... uhm, screwy.
Even if you can succesfully insert something specified as a sequence/identity/call_whatever_you_like you're putting yourself in harm's way because there are additional steps to take to ensure things won't get screwed afterwards, such as trying to insert a row that happens to hold "1234" as the next value whose "place" has been already filled up.
On the top of my head, for example, postgresql needs resyncing of the sequence of the serial, and mssql needs special statements to wrap things up.

<tl;dr> we AREN'T going to support in any way inserting a specific value to a table created by web2py, since we already ruled out that it's a silly thing to do (and, as explained here, there are so much more things to do to support it in a safe and standardized way). 
That being said, we want to be nicer to people who are using legacy tables with an 'id' column that is an integer BUT is not autogenerated. So, ok for the fix, not ok for the test-case.

niphlod:
BTW: I cancelled the appveyor build 'cause mssql test were hanging on the insert ^_^'

ilvalle:
The test is an example of what the user reporting the issue is doing on mysql https://groups.google.com/forum/#!msg/web2py/ipXEtxUbt2w/z_7iDdEuK6wJ
now it works. 
IF you agree, I'll add a skip for mssql so you can re-enable the build and the PR can be merged. 
We can add/rewrite tests afterwards.

niphlod:
the user is dumb. it will lead to errors, and we already explained why for web2py-generated tables. web2py should accomodate for dumbness (only for legacy tables out of web2py's control) but not suggest it nor propose it, and surely not test it in that way. even in the "most respected backend" web2py has, which is postgres, it still leads to inconsistencies (and most surely, errors down the path). 

tl;dr: remove that test alltogether.

ilvalle:
tests removed

mdipierro:
is this ready to merge?

niphlod:
tests are passing so I'd guess it's ready.

PR #196, opened 2015-05-22
Fix MongoDB handling of strings which are not encodable via UTF-8

stephenrauch:
And since we are adding a blob decode, add bytearray handling.


pydal/adapters/mongo.py
```diff
@@ -8,6 +8,8 @@
 from ..helpers.classes import SQLALL, Reference
 from ..helpers.methods import use_common_filters, xorify
 from .base import NoSQLAdapter
+from bson import Binary
```
ilvalle: mind that such imports raise troubles to users that don't have pymongo/bson installed

stephenrauch:
These tests fail when importing the mongodb driver on non-mongodb tests.  When I run the test locally against sqlite, the tests pass.  I am unsure how best to proceed.

PR #218, opened 2015-06-11
finally, appveyor in the org "naming scheme"

niphlod:
BTW: should we update travis.yml to test py3 too ? seems a good time to
start


niphlod:
whoops: I meant 

"should we update travis.yml to report coverage on py3 too ? seems a good time to
start"

gi0baro:
+1 on converage also on py3. Would you update this PR, or should I merge this and you will propose another PR?

niphlod:
merge this in, then I'll do another PR to enable py3 coverage. Need to do some tests on my fork before :D

Issue #73, opened 2015-03-10
MongoDB with versioning and ondelete


franciscocosta:
using CASCADE or SET NULL, it does not matter, web2py does not remove the reference from the referencing collection or delete the reference...

More about this issue here: https://groups.google.com/forum/#!msg/web2py/0tVp5QVR1-8/8F4Xojdug7QJ

PR #219, opened 2015-06-13
MongoDB add support for ondelete CASCADE and SET NULL

stephenrauch:
Resolves https://github.com/web2py/pydal/issues/73 for references.

This does nothing for list:references. Should it?


gi0baro:
Seems that changing the tests causing errors on MSSQL. I think we need @niphlod here.

PR #220, opened 2015-06-15
Fix unicode for postgres adapter

ilvalle:
- added a basic 'unicode' test, not sure if it works on all mssql adapters.

### DIFF
--- a/pydal/adapters/postgres.py
+++ b/pydal/adapters/postgres.py
@@ -332,6 +332,12 @@ class PostgreSQLAdapter(BaseAdapter):
             raise ValueError('Invalid mode: %s' % mode)
         return ['DROP TABLE ' + table.sqlsafe + ' ' + str(mode) + ';']
 
+    def execute(self, *a, **b):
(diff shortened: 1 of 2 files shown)


gi0baro:
Seems that is causing errors on mssql + python3.. maybe @niphlod can help on this?

niphlod:
it'll be hardly solvable. In the end, unicode handling with py2 and "without n adapter" was a de-facto "let the magic happen" at the adapter level. I'll see if it's fixable but for all matter and purposes if it works on "with the n adapter" it's thumbs up.

gi0baro:
@mdipierro what about MSSQL?

mdipierro:
I do not know. I followed niphlod's advice. The number of users mssql+py3 is small for us to worry at this point.

gi0baro:
@mdipierro at least we should skip these new tests on MSSQL to avoid broken status on builds.

mdipierro:
I agree. Can you remove them?

gi0baro:
I will do it later today

PR #221, opened 2015-06-15
py3 coverage

niphlod:
coverage for all but 2.6 thanks to unified coverage on codecov.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
       env: ADAPTER=imap
 
 before_script:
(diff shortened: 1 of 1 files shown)


gi0baro:
Maybe we should run coverage only on 3.4? (as well as we calculate it only on 2.7, not 2.6..)

niphlod:
It doesn't add up,even if hardly it'll grow between the two. Feel free to discard this pr and modify to send coverage only for 3.4

mdipierro:
Discarding as requested.

PR #230, opened 2015-06-22
Use better colnames generation in MongoDBAdapter

gi0baro:
Avoid errors when `str(Field)` doesn't return 'tablename.fieldname'

### DIFF
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import datetime
-import re
 
 from .._globals import IDENTITY
 from .._compat import integer_types, basestring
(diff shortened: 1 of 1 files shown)


stephenrauch:
Hello @gi0baro,

In my testing, this change appears to have broken several test cases. The changed line was originally inserted because testing showed that expression based fields do not return a name with the 'tablename.fieldname' pattern. Do you have a test case that fails that shows the need for this change?

gi0baro:
@stephenrauch you right, I merged it too quickly. Gonna revert this and rethink on a different solution.

stephenrauch:
I would be happy to help. Can you describe the problem, or do you have a failing test case?

gi0baro:
The issue comes when the `Field` is subclassed and doesn't return 'tablename.fieldname' on `__str__` method. The idea was to unify the way `colnames` are generated, since the issue is reproducible only with `MongoDBAdapter` as is the only one calling `str(Field)` to build the column name. Is not really *pyDAL* related, since you will never have this issue using pyDAL, is a *risk* for external implementation, since the behavior should be the same for all the adapters.

PR #232, opened 2015-06-23
MongoDB: Refactor colname generation in select()

stephenrauch:
Help avoid errors when str(Field) doesn't return 'tablename.fieldname'


gi0baro:
@stephenrauch thanks!

PR #242, opened 2015-06-30
MongoDB: Implement expressions in queries for delete

stephenrauch:
Add code and test cases for expressions in update, delete and select queries.
Add more test cases from sql.py to nosql.py and enable for mongo.  Fix bugs for same.
Add mongo specific test cases, especially to exercise mongo driver error paths.

### DIFF
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -1880,7 +1880,7 @@ class NoSQLAdapter(BaseAdapter):
         '''
         return None
 
-    def represent(self, obj, fieldtype):
+    def represent(self, obj, fieldtype, object_id=long):
         field_is_type = fieldtype.startswith
         if isinstance(obj, CALLABLETYPES):
             obj = obj()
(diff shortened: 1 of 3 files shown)


pydal/adapters/base.py
```diff
@@ -1880,7 +1880,7 @@ def execute_test_query(self):
         '''
         return None
 
-    def represent(self, obj, fieldtype):
+    def represent(self, obj, fieldtype, object_id=long):
```
gi0baro: I'm not sure is good to have and `object_id` parameter here which will be used only with Mongo.
Probably I will merge this after **15.07** release and then refactor the code a bit.

gi0baro:
@mdipierro let's wait to merge this until 15.07 release please.

pydal/adapters/base.py
mdipierro: OK. Then I will not merge this.

stephenrauch:
@gi0baro when is 15.07 expected?

Also the change for object_id is pretty tame from the standards I am used to.  Should I avoid making any changes to files other than the driver?  I can certainly code around the base driver, but the bottom line is that then any changes to base will need to be reflected in driver, and this becomes tougher to maintain.

gi0baro:
@stephenrauch 15.07 is expected to be released before the 10th (should be the expiration date for the milestone here on github).
Ideally speaking, the `represent` method should be easier to inherit and/or extend. Considering this (and maybe in the next version I can rethink a bit on how to rewrite it with a better code), you shouldn't add a parameter to the `NoSQLAdapter.represent`, which is in common with other adapters, passing this parameter from `MongoDBAdapter`, but instead, write a different `represent` method into the `MongoDBAdapter` that handle the case you're trying to fix.

stephenrauch:
Removed changes to base.represent()

gi0baro:
Ok, this should close #244. Thanks @stephenrauch

PR #249, opened 2015-07-06
Reverted old behavior on validate methods

### LINKED ISSUE #245: `validate_and_insert` and `validate_and_update` behavior
Continue from https://github.com/web2py/web2py/issues/1013
**I'm considering this blocking for the next release.**

The main issue comes from https://github.com/web2py/pydal/pull/238 where I accidentally changed the behavior of the 2 methods.

@niphlod @ilvalle @mdipierro We should decide if the values in `Field.default` should be injected on the values passed to the method, when they're missing.

Easy solution: update my changes to consider default values and don't break any prior usage of them.

Let's talk.

gi0baro:
Closes #245

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -709,12 +709,17 @@ class Table(Serializable, BasicStorage):
             [f(fields, ret) for f in self._after_insert]
         return ret
 
-    def _validate_fields(self, fields):
+    def _validate_fields(self, fields, defattr='default'):
         response = Row()
         response.id, response.errors = None, Row()
         new_fields = copy.copy(fields)
         for fieldname in self.fields:
-            value, error = self[fieldname].validate(fields.get(fieldname))
+            default = getattr(self[fieldname], defattr)() \
+                if callable(getattr(self[fieldname], defattr)) else \
+                getattr(self[fieldname], defattr)
+            value, error = self[fieldname].validate(
+                fields.get(fieldname), default
+            )
             if error:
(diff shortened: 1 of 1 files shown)


pydal/objects.py
```diff
@@ -709,12 +709,17 @@ def insert(self, **fields):
...
         response = Row()
         response.id, response.errors = None, Row()
         new_fields = copy.copy(fields)
         for fieldname in self.fields:
-            value, error = self[fieldname].validate(fields.get(fieldname))
+            default = getattr(self[fieldname], defattr)() \
+                if callable(getattr(self[fieldname], defattr)) else \
+                getattr(self[fieldname], defattr)
+            value, error = self[fieldname].validate(
+                fields.get(fieldname), default
```
mdipierro: there is a bug here. I will merge and fix.

pydal/objects.py
```diff
@@ -709,12 +709,17 @@ def insert(self, **fields):
...
 
-    def _validate_fields(self, fields):
+    def _validate_fields(self, fields, defattr='default'):
         response = Row()
         response.id, response.errors = None, Row()
         new_fields = copy.copy(fields)
         for fieldname in self.fields:
-            value, error = self[fieldname].validate(fields.get(fieldname))
+            default = getattr(self[fieldname], defattr)() \
+                if callable(getattr(self[fieldname], defattr)) else \
```
mdipierro: this can be optimized. Will fix after merge.

PR #251, opened 2015-07-07
fix mistakes that make in/belongs tests failed

BuhtigithuB:


mdipierro:
I am not convinced we should support this syntax in smart_query:

a_table.integer_field in "1, 2, 3"

I think it should be

a_table.integer_field in (1, 2, 3)

instead.

mdipierro:
We should not support this syntax.

PR #257, opened 2015-07-13
MongoDB: implement select(having)

stephenrauch:
Implement select(having).
Fix bug when attempting to update id field.
Deal with server version 2.4 oddities.

### DIFF
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -374,7 +374,7 @@ class MongoDBAdapter(NoSQLAdapter):
                     # do not update id fields
                     for fieldname in ("_id", "id"):
                         if fieldname in self.field_dicts:
-                            self.field_dicts.delete(fieldname)
+                            del self.field_dicts[fieldname]
             else:
                 if crud == 'update':
                     self._add_all_fields_projection(self.field_dicts)
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -402,8 +402,12 @@ class MongoDBAdapter(NoSQLAdapter):
                 #elif crud == 'delete':
                 #    pass
 
-        from bson.objectid import ObjectId
-        NULL_QUERY = {'_id': {'$gt': ObjectId('000000000000000000000000')}}
+        try:
+            from bson.objectid import ObjectId
+            NULL_QUERY = {'_id': {'$gt': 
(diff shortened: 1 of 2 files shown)


ilvalle:
Mind that this PR contains several commits already included in https://github.com/web2py/pydal/pull/255
In addition you should rebase your branch too, so that one clear commit is provided

stephenrauch:
In general I was trying to submit one commit per feature. Also I had been trying to not submit a PR until the previous PR has been merged. Given the inherent lag for merging, I continue on the same stream after a PR. Then when the previous PR is merged, I submit a new PR with the work done since the previous PR was submitted. I however recently lost my job, so the features were coming quickly.

If I mash features together into one big commit, and there is a problem later, it is harder to tell why a particular change was made. If I submit one branch per feature, then subsequent features will have commits from previous features if those features have not been merged. If the PR is not merged, but there are no merge conflicts, I am expected to go back and do additional work to rebase? Having to do additional work, as well as losing track of the change progression doesn't seem like a good idea.

This PR is built on top of #255, because they overlap the same code. It has been several days since I submitted #255, and it has not been merged or even commented on. Yours is the only comment I have received and it isn't terribly encouraging. Bottom line is that I was moving forward considerably faster then the results were being merged. I cannot see how I should have done this much differently except to have slowed down and waited for merges to happen.

ilvalle:
If the pr takes too much time to be merged you can add a second commit with
the new feature/fix directly in the same pr (#255).

Rebasing is useful to better organize and follow the development of any big
project as pydal. One commit per feature is not mandatory but surelly when
feasible is more than welcome.

gi0baro:
@stephenrauch Sorry for the merge delay of #255 
I think the idea of @ilvalle was intended to be applied now that the other branch is merged in master.

If doesn't bother you:
- Rebase your branch on pydal master
- Push force the rebased branch

The PR will be updated automatically.

stephenrauch:
Is this the desired result?

gi0baro:
nope. Something like this:

```bash
git checkout master
git pull pydalorigin master
git checkout feature
git rebase master
git push yourforkorigin feature --force
```

where *pydalorigin* and *yourforkorigin* are respectively pydal repo and your forked repo

stephenrauch:
Code submission take 17....

PR #264, opened 2015-08-06
added getitem method for json field in postgresql adapter

manuelep:
I've added a getitem method for json field in postgresql adapter that works accordingly to [Postgresql documentation](http://www.postgresql.org/docs/current/static/functions-json.html).
I've used the **#>** operator in order to support getting value from nested objects.


niphlod:
please add tests and implementations for all other backends...

manuelep:
Can you please point me to some example of other similar method?

niphlod:
to add tests, use all the boilerplates found here https://github.com/web2py/pydal/blob/master/tests/sql.py

then, you're the one pushing for a feature into an abstraction layer ... abstract! :P 

IMHO on postgresql getitem() can use its own syntax, while on other backend you should fetch the entire field and select at "python-time" the array element.

mdipierro:
This will be added as soon as tests are in.

mdipierro:
I know this is a feature only supported by postgres at this time and test are not in but I will include it. It encodes some useful knowledge and postgres is backend of choice.

niphlod:
we soooooo need a matrix with supported features for each backend or at this point we'll get lost .

mdipierro:
yes. We could somewhat automate it if we have a method to instantiate adapters even if the corresponding drivers are missing. In fact this would allow us to use pydal to generate SQL for a backend even without that backend.

PR #266, opened 2015-08-07
Allow usage of custom classes for Row (only subclasses)

gi0baro:
Added `DAL.row_class` parameter to allow usage of custom classes for single rows (only as subclass of `Row`).

Note: I've changed only code that return row objects to user, internally pyDAL still using its `Row` class.

### DIFF
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -1644,7 +1644,9 @@ class BaseAdapter(with_metaclass(AdapterMeta, ConnectionPool)):
         """
         Return a parsed row
         """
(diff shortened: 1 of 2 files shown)


gi0baro:
@mdipierro @niphlod @ilvalle feedbacks?

mdipierro:
Can we make a db.Row instead of db.row_class() like we do for db.Field and db.Table?

gi0baro:
@mdipierro yep, I will update this PR

gi0baro:
@mdipierro this is ready to be merged.

gi0baro:
@mdipierro Rebased this on master. If everything ok, please merge it.

PR #276, opened 2015-08-18
version file

ilvalle:


gi0baro:
Why this?

ilvalle:
Can be used in any project using pydal

gi0baro:
Ok, but the build is broken and maybe is better to hardcode the version inside the package

niphlod:
dunno about the requirement of having an external version file, but the read_file implementation is silly since 2.4 was never in scope (actually, I seem to recall that 2.5 was dropped too). There's the much more standard - and elegant - 
    with open('foo/bar', 'r') as f:
           content = f.read()
works in jython and pypy too.

mdipierro:
ok. let's fix the implementation as suggested by niphlod. I do not object to have a VERSION file. It is cleaner.

mdipierro:
What do we do about this?

ilvalle:
Beside the implementation, is it worth to have such version file?

mdipierro:
I think there is value in knowing the version of pydal programmatically.

mdipierro:
Any news on this? I will close this and wait for a new PR.

PR #277, opened 2015-08-18
Added version to init

gi0baro:
Alternative to #276

### DIFF
--- a/pydal/__init__.py
+++ b/pydal/__init__.py
@@ -1,3 +1,5 @@
+__version__ = '15.08.dev0'
+
 from .base import DAL
 from .objects import Field
 from .helpers.classes import SQLCustomType
--- a/setup.py
+++ b/setup.py
@@ -17,10 +17,19 @@ Links
 * `documentation <http://www.web2py.com/books/default/chapter/29/06/the-database-abstraction-layer>`_
 """
 
+import re
+import ast
 from setuptools import setup
+
+_version_re = re.compile(r'__version__\s+=\s+(.*)')
+
+with open('pydal/__init__.py', 'rb') as f:
+    version = str(ast.literal_eval(_version_re.search(
+        f.read().decode('utf-8')).group(1)))
+
 setup(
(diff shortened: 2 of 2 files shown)


gi0baro:
Note: the build is not passing only on gae because we had to upgrade the api version for tests

ilvalle:
Should I close the original PR (https://github.com/web2py/pydal/pull/276) ?

gi0baro:
Dunno, I was waiting for feedbacks about the different approach.

ilvalle:
I think that a VERSION file is more cleaner to those people that browser the project.
For applications using pydal having the version in a python file (as in __init__.py) is easier to handle, although a get_version method that reads VERSION can be provide.

gi0baro:
@ilvalle I think that people that browse the project can use git tags to view a specific version, and locally they downloaded a specific version, or the master with git (so they can use tags again). I think this is just the simplest solution for everybody.

stephenrauch:
I also find a VERSION file much cleaner, and would suggest a get_version() that reads the file.

ilvalle:
@gi0baro you are right if people have downloaded the code using git, otherwise you don't have access to git tags.
Given that, I'd update https://github.com/web2py/pydal/pull/276 to provide a get_version() method using the approach proposed by @niphlod in https://github.com/web2py/pydal/pull/276#issuecomment-132325446

gi0baro:
@ilvalle as the developers prefer. 
With my PR `__version__` is accessible inside the pydal package (`pydal.__version__`), and I kept clean the python package (without adding files != .py in the package).
I don't like adding a VERSION file inside the package, and putting it outside the package is just the same of looking at *setup.py* or the changelog (IMHO).

mdipierro:
I am not merging code that does not pass travis tests. Feel free to discuss on dev group.

gi0baro:
@mdipierro this is failing just because is based on the old GAE version, I can rebase with the latest one.

PR #280, opened 2015-08-23
Inject virtual fields in order on Rows

gi0baro:
Actually virtual fields are injected with a random order caused by an `iteritems()` call on the `Table` object, is not possible to access virtual fields trough other virtual fields (unless you're lucky enough to have the right order).

Since there's no an optimal and simple way to parse dependencies between virtual fields, is allowed the access between them using the definition order by the user.

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -576,10 +576,13 @@ class Table(Serializable, BasicStorage):
                 raise SyntaxError(
                     'value must be a dictionary: %s' % value)
             self.__dict__[str(key)] = value
-            if isinstance(value, FieldVirtual):
-                self._virtual_fields.append(value)
-            elif isinstance(value, FieldMethod):
-                self._virtual_methods.append(value)
+            if isinstance(value, (FieldVirtual, FieldMethod)):
+                if value.name == 'unknown':
+                    value.name = str(key)
+                if isinstance(value, FieldVirtual):
+                    self._virtual_fields.append(value)
+                else:
+                    self._virtual_methods.append(value)
 
     def __setattr__(self, key, value):
         if key[:1] != '_' and key in self:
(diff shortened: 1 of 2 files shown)


gi0baro:
@mdipierro Now pass tests. Feedbacks?

ilvalle:
With ```table._virtual_fields``` and  ```table._virtual_methods``` a few things can be also refactored in a second phase.

pydal/objects.py
```diff
@@ -570,11 +576,18 @@ def __setitem__(self, key, value):
...
+                if isinstance(value, FieldVirtual):
+                    self._virtual_fields.append(value)
+                else:
+                    self._virtual_methods.append(value)
 
     def __setattr__(self, key, value):
-        if key[:1]!='_' and key in self:
+        if key[:1] != '_' and key in self:
             raise SyntaxError('Object exists and cannot be redefined: %s' % key)
-        self.__dict__[key] = value
```
ilvalle: Why?

pydal/objects.py
```diff
@@ -570,11 +576,18 @@ def __setitem__(self, key, value):
                 raise SyntaxError(
                     'value must be a dictionary: %s' % value)
             self.__dict__[str(key)] = value
+            if isinstance(value, (FieldVirtual, FieldMethod)):
```
ilvalle: This should be moved elsewhere.
VirtualFields are defined only during table.define_table as for normal Fields.
After table.define_table neither Fields, nor VirtualFields can be defined. 

If we want so, we should introduce a new method in the table Api, something like:
table.update_fields. Which will check for all new added fields.

gi0baro: I've made this change because in tests we are defining virtual fields just setting them as attributes. I thought it was a feature. If not, we should change tests before.

gi0baro: @ilvalle see https://github.com/web2py/pydal/blob/master/tests/sql.py#L229

ilvalle: Ok I saw that It's documented in the same way in the book

gi0baro: So the current implementation is ok?

PR #291, opened 2015-09-15
Add COLLATE to DAL

timnyborg:
Allows for accent-insensitive DAL queries like:
db.person.firstname.collate('Latin1_General_CI_AI').ilike('Joao')

Only an MSSQL implementation, but the same should work on MySQL, and Postgres will just require quotation marks.


niphlod:
I'd leave it not implemented (there are database default collations or views for that) if it's not implementable AND tested in all other backends.

mdipierro:
Before we include this we should at least make sure it works with sqlite and postgresql which are the most common dbs.

mdipierro:
Thanks for the help but looks like there is not much agreement on this.

PR #307, opened 2015-11-10
Update to SPATIALLIBS

boa-py:
The most recent version of the spatialite DLL is named mod_spatialite.dll - the value of SPATIALLIBS should be modified accordingly.

This proposed edit addresses an old issue posted by Anthony (although his proposed edit mentioned another outdated version of the spatialite DLL): https://github.com/web2py/web2py/issues/1062

### DIFF
--- a/pydal/adapters/sqlite.py
+++ b/pydal/adapters/sqlite.py
@@ -125,7 +125,7 @@ class SQLiteAdapter(BaseAdapter):
 
 
 SPATIALLIBS = {
-    'Windows':'libspatialite',
+    'Windows':'mod_spatialite.dll',
     'Linux':'libspatialite.so',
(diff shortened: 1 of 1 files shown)


mdipierro:
I have not tested this but I trust you, since you are probably the only one to use it. :-)

boa-py:
@mdipierro - thanks, Massimo - appreciate that you trust me on this one, although if anyone wants to verify for themselves, the pre-compiled Windows DLLS are [here](http://www.gaia-gis.it/gaia-sins/). ;)

Sure, the spatialite libs are limited compared to, say, postgres's geospatial features, but they still provide some useful functionality - not sure why more people aren't using spatialite with web2py.

PR #308, opened 2015-11-10
fix 306 parse_datetime() misrepresentation of millisecond part

kmcheung12:


stephenrauch:
Good catch on the length of the fractional seconds.  However, I believe the fix proposed fails if more than 6 digits are given.  I would suggest instead something like:

``` python
if ms and ms[0] == '.':
    ms = int(float('0' + ms) * 1000000)
else:
    ms = 0
```

kmcheung12:
@stephenrauch You are right! I have made the corresponding commit

stephenrauch:
@kmcheung12 

This build looks like it failed because of the addition of:

```python
from .base import *
```

in ```tests/__init__.py```.

I am not sure why you needed to do this.  But, I have just committed a change (https://github.com/web2py/pydal/pull/309) which allows this line to work, by fixing these test cases for Mongo, and by disabling the tests for the other NOSQL databases for which these test cases were failing.

kmcheung12:
@stephenrauch import by mistake..

gi0baro:
@kmcheung12 Please, add a commit to this PR to undo changes on the `tests/__init__.py`

kmcheung12:
I have undone the changes on `tests/__init__.py` but the build still fail. Can someone kindly advice?

stephenrauch:
It looks like the build error was actually due to some glitch in AppVeyor, not because of a code problem. Since this is small change with several submits at this point, you might consider using 'git rebase' to squash the changes into a single commit. If you then push that to github, the test build should run again.

kmcheung12:
Feel free to have a look

PR #318, opened 2015-12-24
Compute fields are always computed

cassiobotaro:
Related to https://github.com/web2py/web2py/issues/1128.


mdipierro:
There is an issue with your fix. it should be possible to pass a value for a computed field and have it not-computed. The problem is actually a bug somewhere else. Fixed here: 81d9453

Issue #296, opened 2015-09-25
MSSQL3Adapter & MSSQL3NAdapter combination of orderby and limitby resulting in a wrong sql query


skoa-h:
```python
db.define_table('tag', Field('title', 'string'),)

tags = db(db.tag.id > 0).select(
					  db.tag.id,
					  db.tag.title,
					  orderby=db.tag.title|db.tag.id,
					  limitby=(1,40),
					  )
					  
print db._lastsql
```
```sql
SELECT  f_0, f_1 FROM (SELECT  ROW_NUMBER() OVER (ORDER BY tag.id, tag.title) AS w_row, tag.id AS f_0,  tag.title AS f_1 FROM tag WHERE (tag.id > 0) ) TMP WHERE w_row BETWEEN 2 AND 40;
```
in OVER (ORDER BY **tag.id**, **tag.title**) the order of the statements is wrong, it seems id is always first.
I found the bug because SQLFORM.grid messed up the column sorting in pages > 1.

skoa-h:
The problem seems to be https://github.com/web2py/pydal/blob/master/pydal/adapters/mssql.py#L343
```python
return 'SELECT %s %s FROM (SELECT %s ROW_NUMBER() OVER (ORDER BY %s) AS w_row, %s FROM %s%s%s) TMP WHERE w_row BETWEEN %i AND %s;' % (sql_s,sql_f_oproxy,sql_s,sql_f,sql_f_iproxy,sql_t,sql_w,sql_g_inner,lmin,lmax)
```
sql_f is given for the OVER (ORDERBY %s), this should be sql_o_inner, which is already prepared in https://github.com/web2py/pydal/blob/master/pydal/adapters/mssql.py#L336

Proposed fix:
```python
return 'SELECT %s %s FROM (SELECT %s ROW_NUMBER() OVER (ORDER BY %s) AS w_row, %s FROM %s%s%s) TMP WHERE w_row BETWEEN %i AND %s;' % (sql_s,sql_f_oproxy,sql_s,sql_o_inner,sql_f_iproxy,sql_t,sql_w,sql_g_inner,lmin,lmax)
```

niphlod:
this can be integrated, although if you're on MSSQL >= 2012 it's best to use mssql4:// . @gi0baro : do you want a PR or can you fix it directly ?

gi0baro:
@niphlod Will fix asap.

niphlod:
[...]

PR #332, opened 2016-02-28
Refactoring

gi0baro:
As introduced on the group (https://groups.google.com/forum/#!topic/web2py-developers/zw6-bFpe96s) this is a major refactoring of the code base.

Feedbacks and comments are welcome.

Intentions and the rationale:
- Stop using the BaseAdapter as a super-collector (sometimes taking place of SQLite – ed. note)
- Better using inheritance between classes
- Simplify the writing of new adapters (and increase the contributions)
- De-uglify a lot of the code making it simpler to understand and debug
- Allowing to easily implement performance optimizations

Refactoring steps:
- [x] BaseAdapter re-structuring
- [x] Better SQLAdapter and NoSQLAdapter
- [x] Base *dialects* structure
- [x] Base *parsers* structure
- [x] Base *representers* structure
- [x] Migrations separation from base adapter
- [x] SQLite and Spatialite adapters
- [x] Postgresql adapter
- [x] Postgresql 2/new adapter
- [x] MySQL adapter
- [x] MSSQL adapter
- [x] *jdbc* versions of adapters
- [x] Review of NoSQLAdapter
- [x] MongoDB Adapter
- [x] Other SQL adapters
- [x] Other NoSQL adapters
- [x] Final code cleaning

Changes from 16.03:
- Postgre adapter no longer parse datetime objects
- `MSSQL2Adapter` no longer exists since was the same of `MSSQLNAdapter` (but the uri is still valid)
- Includes fix for #296 
- `BaseAdapter.connection` and `BaseAdapter.cursor` are now thread safe
- Removed `_lastsql` in adapters since not thread safe
- Empty values for `ObjectId` *fields* in `MongoAdapter` now are stored (and returned) as `None` instead of a fake `ObjectId` converted to integer `0`


mdipierro:
ready to be merged?

niphlod:
IMHO he didn't say it's ready ... and btw lots of tests are failing, which are expected until everything gets "tuned" to the new structure.

gi0baro:
@mdipierro @niphlod No, I said is not. It's actually missing all the adapters except for sqlite.
The checklist should give you an idea of the status.

In the meantime, would be helpful to have feedbacks on the new parts (dialect, parser, representer).

gi0baro:
@niphlod I've got some SQL syntax errors with MSSQL: https://ci.appveyor.com/project/web2py/pydal/build/1.0.191/job/4hign5cdur2qtoih

Do you have any hints?

niphlod:
it'll be helpful to see the underlying generated query  ;-)

gi0baro:
@niphlod the bad query should be this one:

```sql
CREATE TABLE staff(
    id INT IDENTITY PRIMARY KEY,
    name VARCHAR(512) NULL,
    food VARCHAR(512) NULL,
    tvshow INT  NULL , CONSTRAINT staff_tvshow__constraint FOREIGN KEY (tvshow) REFERENCES tvshow (id) ON DELETE CASCADE
);
```

niphlod:
nope, that statement is perfectly valid...

gi0baro:
@niphlod The only other table defined in the test produces this sql statement:

```sql
CREATE TABLE tvshow(
    id INT IDENTITY PRIMARY KEY,
    name VARCHAR(512) NULL,
    rating DOUBLE NULL
);
```

niphlod:
tested locally. the problem is that the "DOUBLE" datatype doesn't exist in mssql

gi0baro:
@niphlod thank you, gonna fix it.

gi0baro:
@niphlod I'm gonna ask for your help again about this: https://ci.appveyor.com/project/web2py/pydal/build/1.0.194/job/s8p04if0kcanaamx

I'm just getting mad trying to figure out which part of the code is altering the unicode representation (or parsing?)..

gi0baro:
If I print the sql string looks like this:

```sql
INSERT INTO a(a) VALUES (N'Aé A');
```

niphlod:
uhm. bit puzzled because I don't recognize the original test (maybe a commented representation would have been helpful). is ucs really a sequence of
- latin capital letter A
- latin capital letter A with tilde
- copyright sign
- space
- latin capital letter A 

?

niphlod:
it seems that tables are not created with the right syntax: string fields are created as VARCHAR(512) instead of NVARCHAR(512) ... and I don't understand why: it seems that dialects are registering correctly but not field types for dialects.

niphlod:
ok, duplicating the register for string fields corrects the issue. Maybe there is a subtle bug in dialects that inherits from multiple classes. Dunno the exact reason though.

Here's the deal: it's _compat.py's "to_unicode" that does ucs.encode('utf-8') and translates to u'A\xe9 A' . all that it'd take would be to test what to_unicode did before this PR on that expansion. sorry but I have limited time ^_^

willimoa:
Do all adaptors need refactoring to align with the new BaseAdapter ?    I've been working on the Teradata Adapter, what needs to be done ?  Is the new BaseAdapter available ?

gi0baro:
@niphlod thank you, will re-check the flow.
@willimoa yes, all the adapters need to be refactored. Teradata is not yet ported, will keep you updated.

gi0baro:
@mdipierro @niphlod I have some troubles with google datastore adapter. Right now this test fails:

```python
  File "tests/nosql.py", line 1609, in testRun
    assert isinstance(db4(db4.tvshow).select().first().id, long) == True
AssertionError
```

But I don't get why it should pass, since also in the old adapter I see:

```python
def parse_id(self, value, field_type):
    return value
```

Any ideas?

niphlod:
hate GAE.... sure that it got executed and passed in the old suite ?

gi0baro:
@niphlod I hate it too, but is the latest adapter to finish this PR.

Yes, I have not modified the tests, and the last suite run on the master reports:

```
testRun (tests.nosql.TestContains) ... skipped 'Contains not supported on GAE Datastore. TODO: IMAP tests'
testRun (tests.nosql.TestDALDictImportExport) ... ok
```

while the PR:

```
testRun (tests.nosql.TestContains) ... skipped 'Contains not supported on GAE Datastore. TODO: IMAP tests'
testRun (tests.nosql.TestDALDictImportExport) ... FAIL
```

:/

mdipierro:
In the pervious implementation we always made sure that the id was a long at the web2py level. There are functions in there that convert other ids to long and vice versa. We do the same in mongo. If the id is not a long sqlform breaks.

gi0baro:
@mdipierro yes, but my question was "which part of the existing code did that?" since the parser just get the value from the database: https://github.com/web2py/pydal/blob/master/pydal/adapters/google_adapters.py#L157#L158

Maybe I wrongly ported the representation? is pyDAL that should store the long inside the db?

mdipierro:
I need to double check this. Will do it later today.

gi0baro:
@mdipierro did you find anything?

mdipierro:
Found something but not satisfied. If you grep parse_id in adapters you find the function (in mongo) that converts the id to long. In Gae that function does not perform any conversion. So the question is, should it? This worked before on gae (the id was a long), so what changed? I will run a few more tests and ask some gae users.

gi0baro:
@mdipierro do you have any updates? I still don't understand how it worked before.

In fact, even the `insert` method convert the key to long:
```python
def insert(self, table, fields):
        dfields = dict((f.name, self.represent(v, f.type)) for f, v in fields)
        tmp = table._tableobj(**dfields)
        tmp.put()
        key = tmp.key
        rid = Reference(key.id())
        rid._table, rid._record, rid._gaekey = table, None, key
        return rid
```

So I would expect the parser to convert the value, not to return it untouched. I'm missing something.

gi0baro:
@mdipierro got it. I removed some code from the base adapter. Will complete the PR in the next days.

gi0baro:
@mdipierro @niphlod Finally: this is ready.

@mdipierro Do you want to merge it?

mdipierro:
I will review it during the week-end. Look forward to merge it.

mdipierro:
There are so many changes that this is going to be hard to test, especially for GAE. I will run my own tests but also I will make a new build of web2py so people can check with their existing code.

gi0baro:
@mdipierro I know, but developers are waiting this to push their changes. I'm also waiting on this for indexes and other stuffs. Let's merge this, integrate with web2py, publish an alpha and see if users experience any trouble. I will fix the code asap on case of bugs.

mdipierro:
yes. Will merge today.

mdipierro:
Merging this now but reserve the right to undo if major tests fail.

mdipierro:
I merged but it fails lots of tests: https://travis-ci.org/web2py/web2py/jobs/124079775
I thought this was tested already. Let me know asap how you want to handle this because we cannot git master broken in this way.

mdipierro:
Looks like the main issue is that DAL(None) is not longer working and therefore SQLFORM.factory breaks. The new form.py will not use DAL(None) but we need to keep if for backward compatibility. DAL(None) create a fake dal object that does not establish any connection.

gi0baro:
@mdipierro ok, will fix everything today

PR #349, opened 2016-04-21
Fixed #339

### LINKED ISSUE #339: coalesce() incorrectly expands constant values
When you pass string constant into coalesce, it will be expanded as identifier instead of string constant:
`db().select(db.table.str_field.coalesce('foo'))`
would expand into this:
`SELECT COALESCE(table.str_field,foo) FROM table`
but the expected behavior is this:
`SELECT COALESCE(table.str_field,'foo') FROM table`

gi0baro:


mdipierro:
looks good to me. can I merge?

gi0baro:
@mdipierro yes!

PR #351, opened 2016-04-22
Added per-dialect expressions

gi0baro:
This is a new concept, since allows every dialect to add or customise expressions directly on `Expression` objects.

This PR also add some expressions on `Postgre`. 
@mdipierro @niphlod feedbacks are welcome.


mdipierro:
I like this. Anybody has any objection? Speak now.

PR #358, opened 2016-05-09
Fix #353

### LINKED ISSUE #353: Table.truncate fails
Table.truncate() has a default parameter `mode=None`, which overrides the default of `''` in SQLAdapter.truncate(), and hence produces a query like:

    TRUNCATE TABLE TablenameNone

i.e. appending the `None` to the tablename. Of course that leads to a 

    ProgrammingError: (1146, "Table 'TablenameNone' doesn't exist")

Shouldn't the default in Table.truncate() be `''` as well then? Or does one have to specify a `mode=''` explicitly for Table.truncate()?

gi0baro:


gi0baro:
@mdipierro seems that codecov is broken. This is good for merge.

niphlod:
Codecov is now requiring a yaml config file to be tuned...didn't have any
time to read the docs yet

PR #360, opened 2016-05-10
Enabling entity quoting as default behavior

gi0baro:
This just change pydal behaviour to enable entity quoting per-default.

@mdipierro @niphlod @ilvalle tests are passing, if you have any concerns speak now. This *should* be preferred since avoid a lot of troubles with reserved names for example.

### DIFF
--- a/pydal/adapters/__init__.py
+++ b/pydal/adapters/__init__.py
@@ -33,8 +33,7 @@ class AdapterMeta(type):
             'uploads_in_blob', cls.uploads_in_blob)
         cls.uploads_in_blob = uploads_in_blob
 
-        ## TODO: avoid entity quoting disable
-        entity_quoting = kwargs.get('entity_quoting', False)
+        entity_quoting = kwargs.get('entity_quoting', True)
         if 'entity_quoting' in kwargs:
             del kwargs['entity_quoting']
 


michele-comitini:
I suppose we went through this same path sometimes ago.  It was decided not
to make  it default because some backends poorly manage upper/lowercase
entities.  Also check that no problem raises with Field('Fname') vs
Field('fname') .  In theory both are legit, but some backends make
confusion.

2016-05-10 2:11 GMT+02:00 Codecov <notifications@github.com>:

gi0baro:
@michele-comitini I'm sorry Michele, just forgot who spoken in the past. Do you have any hints on which backends you think will cause troubles?

Also, I've spoken to early. Seems the build is not passing on appveyor, but I don't understand the results of the tests. Seems unrelated? @niphlod do you have any hints?

gi0baro:
Ok, I've missed the edits by @mdipierro on csv imports. Those causing this PR failure.
@mdipierro watch out, current master is broken: https://travis-ci.org/web2py/pydal/builds/129006733

niphlod:
BTW, the previous "discussion" is basically a simple thing: DAL supports a whole lot of backend but, really, it just supports as stable the ones it's unittested on. 
The path we chose was about "sponsoring" the status quo... "if it works, don't touch". But nothing changed since.
Work on several "ephemeral" adapters stopped a long time ago and I'm sure that nobody is testing those. I'm +1 on pushing the feature as it is because I WANT to see users whining about their firebird adapter not working. At some point we'll need to drop adapters that are used by 1% of people and are unmaintained and are not unittested.
BTW: the only real issue, as @michele-comitini suggested, is that mysql and case-sensitivity support is... uhm... flaky. Bonus points for having yet another reason to flag it as a troublemaker rather than a backend.

PR #361, opened 2016-05-11
[WIP] Indexes support

gi0baro:
**Don't merge this until I say is ready**

This implements basic indexes support as requested in #152.
Intended workflow:

```python
# just fields
db.mytable.create_index('index_name', db.mytable.myfield_a, db.mytable.myfield_b)
# with expressions
db.mytable.create_index('index_name', db.mytable.myfield_a, db.mytable.myfield_b.coalesce(None))
# with additional parameters (where is supported only by postgres)
db.mytable.create_index('index_name', db.mytable.myfield_a, where=(db.mytable.myfield_b == False), unique=True)

db.mytable.drop_index('index_name')
```

@niphlod @mdipierro @ilvalle @michele-comitini feedbacks are welcome.

### DIFF
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -2,6 +2,7 @@ import copy
 import sys
 import types
 from collections import defaultdict
+from contextlib import contextmanager
 from .._compat import PY2, with_metaclass, iterkeys, iteritems, hashlib_md5, \
     integer_types
 from .._globals import IDENTITY
@@ -42,6 +43,7 @@ class BaseAdapter(with_metaclass(AdapterMeta, ConnectionPool)):
         self.credential_decoder = credential_decoder
         self.driver_args = driver_args
         self.adapter_args = adapter_args
+        self.expand = self._expand
         self._after_connection = after_connection
         self.connection = None
         if do_connect:
@@ -150,7 +152,7 @@ class BaseAdapter(with_metaclass(AdapterMeta, ConnectionPool)):
                         query = query & newquery
         return query
 
-    def expand(self, expression, field_type=None, colnames=False):
+    def _expand(self, expression, field_type=None, colnames=False):
         return str(expression)
 
     def expand_all(self, fields, tablenames):
@@ -418,7 +420,7 @@ class SQLAdapter(BaseAdapter):
             handler.after_execute(command)
         return rv
 
-    def expand(self, expression, field_type=None, colnames=False):
+    def _expand(self, expression, field_type=None, colnames=False):
         if isinstance(expression, Field):
             et = expression.table
             if not colnames:
@@ -458,6 +460,17 @@ class SQLAdapter(BaseAdapter):
         else:
             rv = expression
         return str(rv)
+
+    def _expand_for_index(self, expression, field_type=None, colnames=False):
+        if isinstance(expression, Field):
+            return expression._rname or self.dialect.quote(expression.name)
+        return self._expand(expression, field_type, colnames)
+
+    @contextmanager
+    def index_expander(self):
+        self.expand = self._expand_for_index
+        yield
+        self.expand = self._expand
 
     def lastrowid(self, table):
         return self.cursor.lastrowid
@@ -802,13 +815,25 @@ class SQLAdapter(BaseAdapter):
             for field in fields]
         sql = self.dialect.create_index(
             index_name, table, expressions, **kwargs)
-        self.execute(sql)
-        self.commit()
+        try:
+            self.execute(sql)
+            self.commit()
+        except Exception as e:
+            self.rollback()
+            err = 'Error creating index %s\n  Driver error: %s\n' + \
+                '  SQL instruction: %s'
+            raise RuntimeError(err % (index_name, str(e), sql))
         return True
 
     def drop_index(self, table, index_name):
         sql = self.dialect.drop_index(index_name, table)
-        self.execute(sql)
+        try:
+            self.execute(sql)
+            self.commit()
+        except Exception as e:
+            self.rollback()
+            err = 'Error dropping index %s\n  Driver error: %s'
+            raise RuntimeError(err % (index_name, str(e)))
         return True
 
     def distributed_transaction_begin(self, key):
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -804,10 +804,12 @@ class SQLAdapter(BaseAdapter):
             index_name, table, expressions, **kwargs)
         self.execute(sql)
         self.commit()
+        return True
 
     def drop_index(self, table, index_name):
         sql = self.dialect.drop_index(index_name, table)
         self.execute(sql)
+        return True
 
     def distributed_transaction_begin(self, key):
         pass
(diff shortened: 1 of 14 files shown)


niphlod:
I'm continuing to not like at all having some features only available for a
backend. It has to stop somewhere or it'll blow in our faces. Moreover,
there is no place in which a support matrix tracks what works in which
backend. Pydal is becoming at each release a PAL rather than a DAL.

That being said it needs a sleuth of tests.

niphlod:
Also, where is the information about indexes stored ? (i.e. is this supposed to be a statement like define_table that is idempotent or a one-off execution ?)

gi0baro:
@niphlod 
1) This is a first-step implementation that doesn't include the migration stuffs, as we agreed on https://groups.google.com/forum/#!searchin/web2py-developers/indexes/web2py-developers/-ovdFU3-rw4/LWPSmOuOAwAJ
Plus, right now I don't have time to re-write the migrator, so this PR embrace the philosophy: *now is better than perfect*.

2) I can't see your point about Postgre, since the only difference here is a `where` parameter in the index creation. In addition, this is not a decision of giving different features, is a different support from the backends and I cannot agree on don't allowing this on postgre because 95% of the backends we support don't have this. And I don't agree on the opposite philosophy as well: just think we have a `for_update` parameter in the `select` method of all adapters just because one of them need it.
That said, I don't know which other features are making DAL a PAL, but it's frankly normal I'm pushing things on Postgre since is the backend I personally use for my projects and in my company, and I code what I need first of all, I don't know which is the syntax of a single command on the majority of backends "supported" by pyDAL, because I don't use them. Not allowing me and my company using postgre features with pyDAL and going back to `execute_sql` or writing raw SQL strings is not an option and would lead me and my company to use something else like SQLAlchemy. And I don't want that.

3) Can we make a list of things we want to test on indexes? I just ended up with the 3 I wrote.

gi0baro:
@niphlod side note to point 2: also SQLAlchemy handle the where option on postgre http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#postgresql-specific-index-options

niphlod:
sidenote to the sidenote, "yep, but it's fully documented" and btw partial indexes are surely supported in MSSQL

without A decision (whatever it might be) towards supporting "sparse" features (and one thing is to support them in SQLite, mysql, postgresql, oracle, mssql and another thing is just supporting them in a SINGLE backend) with "now is better than perfect" attitude is what bit us before (hard). 
If the path of "let's implements bits here and there" is taken, IMHO the lack of documentation about it is astounding.
"here" is this bit, extracting century from date is another bit, supporting inserts with RETURNING is another and bit by bit branches depart from the main tree. See what happened "basing" the whole API on SQLite (as it was originally)...
There are multiple shades of statements:
-  "I don't want pydal to support the common denominator of features of every supported backend, because it will be poor"
- "I want to support every feature backed by the "usual suspects" backends"
-  "I play with postgresql so I don't have the time to implement in other backends"
-  "I don't give a damn about experimental backends"
-  "I won't have time to support anything past postgresql which is my de-facto backend"

I'm just stating that the more things we add to a single adapter (lately postgresql) the more 6 months from now we'll face "why is this not working" questions.

gi0baro:
@niphlod ok, so if I understood correctly, it's just a documentation problem.
I haven't documented the code since web2py uses the book, and I implicitly taught we will update it. If I have to add the doc inline to use rst, let's decide that, but I also expect everything else should be documented in the same way.

Regarding partial indexes on MSSQL, I expect someone more prepared to update the dialect for this functionality after this PR is completed.

niphlod:
No prob. The lack of documentation partly is an issue for anyone running into a "but that worked before" or "but that worked on that backend" and partly for who wants to contribute to reduce the differences with their preferred backend. Also, the API should be marked as experimental if it's not supported by "the usual suspects", just because it may happen to suffer reworkings due to backend limitations (e.g. commit after ddl or not). But I'm digressing.
As no decisions have ever been made, but since we're adults comprehending that not documenting lead to outstanding issues, all that it takes (at least IMHO) is to address a "hole" proceeding forward. Documenting with a matrix on the book or a section on a readme or whatever is just another decision to make.

gi0baro:
@niphlod @mdipierro so, concluding: what else we want/need to consider this ready?

niphlod:
it's "already ready". apart from the aforementioned docs, and tests.

gi0baro:
@niphlod yes, but the question includes tests: what we should test in addition to the ones I wrote?

niphlod:
The only one I can think of since migrations aren't on the table is other backends and index on top of index creation (i.e. to return a meaningful error)

gi0baro:
@niphlod ok, I'm gonna need some help here. I can't get why sqlite tests keeps failing https://travis-ci.org/gi0baro/pydal/jobs/131262831 with this:

```
RuntimeError: Error creating index idx_aa_and_bb
  Driver error: near "(": syntax error
  SQL instruction: CREATE INDEX "idx_aa_and_bb" ON "tt" ("aa",COALESCE("bb",NULL));
```

while if I run this both on python2 and 3 on my machine everything works:

```python
>>> from pydal import DAL, Field
>>> db = DAL(entity_quoting=True)
>>> db.define_table('tt', Field('aa'), Field('bb', 'datetime'))
<Table tt (id, aa, bb)>
>>> db.tt.create_index('idx_aa_and_bb', db.tt.aa, db.tt.bb.coalesce(None))
True
>>> db.tt.drop_index('idx_aa_and_bb')
True
```

any ideas?

niphlod:
after some cold-stoned hits to the wall.... 

"""
Expressions in an index may only refer to columns in the table that is being indexed. Indexes on expression will not work with versions of SQLite prior to version 3.9.0.
"""

python ships with 3.6.21 (or check your own, or Travis-CI, using sqlite3.sqlite_version_info)

niphlod:
BTW, sorry it took too much. Reverting to such an out-of-band thinking made me both mad to have figured it out so late and proud to have remembered why I always update sqlite3 with pysqlite2 everywhere I can.

gi0baro:
@niphlod thank you! Should we skip the test or try to update the version on travis?

niphlod:
Once upon a time tox forced pysqlite2 as a dep. Dunno why it was removed.
Il 23 mag 2016 10:53 PM, "Giovanni Barillari" <notifications@github.com> ha
scritto:

gi0baro:
@niphlod I think I will skip the test on sqlite since we can't know in advance which version of sqlite is used by the user.

gi0baro:
@niphlod I've just rebased the PR on current master, and got back to expressions tests and where conditions test only on postgre, since what we said about SQLite, mySQL not supporting them, and MSSQL reporting syntax errors too: https://ci.appveyor.com/project/web2py/pydal/build/1.0.323/job/5oshrahhv3h9cluj

I think right now this is ok.

niphlod:
yep, MSSQL has filtered (aka "partial") indexes but not indexes on "expressions". I think we just miss the aforementioned "support matrix" and we're good to go.
@mdipierro : thoughts ?

mdipierro:
This is great! Thanks for your hard work.

PR #367, opened 2016-05-19
Allow custom Rows classes

gi0baro:


gi0baro:
@mdipierro I don't know why codecov keeps complaining, but this is good to be merged.

PR #370, opened 2016-05-31
turn off codecov/changes report

niphlod:
hopefully :)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -7,9 +7,7 @@ coverage:
   range: "70...100"
 
   status:
-    changes:
-      default:
-        enabled : no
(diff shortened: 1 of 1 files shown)


gi0baro:
@niphlod if this should disable changes report, then the comment above seems saying this is not working :/

niphlod:
baby steps :-P . It doesn't create the "check" (so it now shows "all checks have passed"). 
To avoid posting the piece in the comment, the layout line should be layout: "header, diff" . That being said, I'm in contact to know why those numbers are being reported in the first place (nothing effectively changed here), because it seems that the commit can't be properly fetched.

PR #397, opened 2016-07-20
fixes #396

### LINKED ISSUE #396: upgrading web2py breaks use of executesql
Before i could: 

```python

db.define_table('top_interrupts_by_frequency',
                Field('activity'),
                migrate=False,  # for a query
                primarykey=['activity']
                )

def top_interrupts_by_frequency(user_id, count=8, min_duration=2):
    dal_set = db.executesql('''
                select activity
                  from timetable
                 where client = ?
                   and ended is not null
                  and cast((julianday(ended) - julianday(started))*24*60*60  as integer) > ?
              group by activity
              order by count(activity) desc
                 limit {}
             '''.format(count), placeholders=(str(user_id), min_duration), fields=db.top_interrupts_by_frequency)
    return dal_set 
```

and in the controller have something like: 

```python
    top_interrupts = top_interrupts_by_frequency(auth.user_id)
    for row in top_interrupts:
        button = A('Int ', STRONG(row.activity), _class='btn btn-success btn-sm btn-outline',_role='button',_href=URL(f='cmd', vars=dict(command='i "{}"'.format(xmlescape(row.activity)))))
```

After upgrading (to 2.14.6-stable+timestamp.2016.05.09.19.18.48) this throws something ugly at my face: 

```text
  File "/home/www-data/web2py/gluon/restricted.py", line 220, in restricted
    exec(ccode, environment)
  File "/home/www-data/web2py/applications/timed/controllers/default.py", line 250, in <module>
  File "/home/www-data/web2py/gluon/globals.py", line 405, in <lambda>
    self._caller = lambda f: f()
  File "/home/www-data/web2py/gluon/tools.py", line 4297, in f
    return action(*a, **b)
  File "/home/www-data/web2py/applications/timed/controllers/default.py", line 119, in index
    button = A('Begin ',STRONG(row.activity), _class='btn btn-default btn-sm btn-outline',_role='button',_href=URL(f='cmd', vars=dict(command='on "{}"'.format(xmlescape(row.activity)))))
  File "/home/www-data/web2py/gluon/packages/dal/pydal/objects.py", line 91, in __getattr__
    raise AttributeError
AttributeError
```
with for the `__getattr__` call: 
```text
Function argument list
(self=<Row {'_extra': {'top_interrupts_by_frequency.activity': 'brk:Break'}}>, k='activity')
```

Trying to get any data out a record requires something like this: 
```python
row['top_interrupts_by_frequency.activity']
```
which you can imagine is quite unlike 
```python
row.activity
``` 
which i have used quite a bit. 
I've updated web2py to the latest, so that's the dal version used right now. I don't know exactly what version i came from. 

edit: new version of web2py added

niphlod:


niphlod:
thanks to @gi0baro for the assistance

PR #400, opened 2016-07-27
Fix #399 - postgres lastrowid select lastval + some PEP8 enhanchements

### LINKED ISSUE #399: lastrowid() for postgres adapter use lastval() which cause issue in case of multiples callback
For more explanation see the related thread : https://groups.google.com/forum/#!topic/web2py/rqkRnsPUpD4

I fix my issue with this mod (line 179 of gluon/packages/dal/pydal/adapters/postgres.py) : 

```python
    # def lastrowid(self, table=None):
    #     if self._last_insert:
    #         return int(self.cursor.fetchone()[0])
    #     else:
    #         self.execute("select lastval()")
    #         return int(self.cursor.fetchone()[0])

    def lastrowid(self, table):
        sequence_name = table._sequence_name
        self.execute("SELECT currval('%s');" % sequence_name)
        return long(self.cursor.fetchone()[0])
```

I don't see why it could cause backward compatibility issue...

I send a PR, in a minute...

BuhtigithuB:


BuhtigithuB:
Travis failed, but it not seems to be caused by the change I made...

BuhtigithuB:
I am gonna try with this :

```python
    def lastrowid(self, table=None):
        if table:
            sequence_name = table._sequence_name
            self.execute("SELECT currval('%s');" % sequence_name)
            return long(self.cursor.fetchone()[0])
        elif self._last_insert:
            return long(self.cursor.fetchone()[0])
```

My problem seems to come from the lines below from the previous implementation :
```python
    if self._last_insert:
        return int(self.cursor.fetchone()[0])
```

I guess this "if" is there to avoid a select every time... But what is the rational?

This is not the case for MSSQL and Oracle adapter, I mean they make a select each time...

Does droping the "if" cause any performance issue?

I also notice that postgres adapter lastrowid() has table=None... Not the other adapter... Could lastrowid() be called without table argument? Was that the reason for the "if"?

Should we make sure that we always pass the table argument?

BuhtigithuB:
quoting issue...

pydal/adapters/postgres.py
```diff
@@ -107,10 +107,10 @@ def after_connection(self):
...
 
     def lastrowid(self, table=None):
-        if self._last_insert:
-            return int(self.cursor.fetchone()[0])
-        self.execute("select lastval()")
-        return int(self.cursor.fetchone()[0])
+        if table:
+            sequence_name = table._sequence_name
+            self.execute('SELECT currval(\'%s\');' % sequence_name)
+            return long(self.cursor.fetchone()[0])
```
mdipierro: This cannot be. It must return something

BuhtigithuB:
you are right... I was trying to solve the quoting issue... Is it safe to consider that table is always provided? all the usage seems to provide table... And all the other adapter have table argument which can't be None...

BuhtigithuB:
It failed like this : 
LINE 1: SELECT currval('""this is table t1"_id_seq"');

Don't know how to resolve this...

This work just fine when I use it : 
```python
    def lastrowid(self, table):
        sequence_name = table._sequence_name
        self.execute('SELECT currval(%s);' % sequence_name)
        return long(self.cursor.fetchone()[0])
```

But tests fails... Maybe something related to _adapt.py _quote()??

Richard

BuhtigithuB:
I mean this : 
```python
def lastrowid(self, table):
        sequence_name = table._sequence_name
        self.execute('SELECT currval(\'%s\');' % sequence_name)
        return long(self.cursor.fetchone()[0])
```

BuhtigithuB:
Some old... I don't understand the quoting issue as tests execute properly here (locally) and when I apply the changes and try it with 2.14.6 it works fine... Any ideas? AppVeyor pass, is there Travis issue?

mdipierro:
Will review asap

mdipierro:
Manually merged.

PR #401, opened 2016-08-01
portalocker: removed pywin32 dep, added tests

niphlod:


mdipierro:
I will close the issue on web2py side too! Than you Simone.

Issue #403, opened 2016-08-07
UnicodeDecodeError when creating new group


matclab:
_Seen with pydal v16.03._

In web2py, when a new user is registering, a new group is created with a localised group name. for example, in french : 'Groupe unique affecté à l'utilisateur 6369470859706368'.

Some excerpt of a debugging session are showned here after:
```
/home/clabaut/web/web2py-lost.re/gluon/packages/dal/pydal/objects.py(726)insert()
-> ret = self._db._adapter.insert(self, self._listify(fields))
(Pdb) p fields
{'role': 'user_6369470859706368', 'description': <lazyT 'Group uniquely assigned to user %(id)s'>}
…
/home/clabaut/web/web2py-lost.re/gluon/packages/dal/pydal/adapters/base.py(1937)represent()
-> obj = to_unicode(obj)
(Pdb) p obj
<lazyT 'Group uniquely assigned to user %(id)s'>
(Pdb) unicode(obj)
(Pdb) str(obj)
"Groupe unique affect\xc3\xa9 \xc3\xa0 l'utilisateur 6369470859706368"
```

This seems similar to a web2py bug corrected in [web2py PR 1163](https://github.com/web2py/web2py/pull/1163).

Using `str` instead of `unicode` everytime or upon `lazyT` objects may be the solution.

matclab:
We should perhaps call `obj.decode` for any object that has a callable `decode` argument (it would apply to `bytes` and `lazyT`.

Any advice appreciated…

matclab:
The following patch might do the trick:
```diff
diff --git a/pydal/_compat.py b/pydal/_compat.py
index 97e9194..d0efd64 100644
--- a/pydal/_compat.py
+++ b/pydal/_compat.py
@@ -86,9 +86,10 @@ def with_metaclass(meta, *bases):
 def to_unicode(obj, charset='utf-8', errors='strict'):
     if obj is None:
         return None
-    if not isinstance(obj, bytes):
+    try:
+        return obj.decode(charset, errors)
+    except AttributeError:
         return text_type(obj)
-    return obj.decode(charset, errors)

```

gi0baro:
I don't agree on this. `lazyT` is a web2py object, the correct methods implementation for Unicode handling should be done in web2py. Patching pydal's compat module like this means only slowing down the whole source code every time `to_unicode` is invoked.

matclab:
@gi0baro, I agree that we shall not do specific things for web2py objects.

[...]

PR #404, opened 2016-08-09
to_unicode: try to use decode method before falling_back to text_type

matclab:
This generalized the case of `bytes` for object that support `decode` method.

It seems to solve the issue #403.


gi0baro:
I'm sorry but I have to reject this, since is breaking python3 support. See also #403 discussion.

PR #408, opened 2016-08-23
Joinable subselects

nextghost:
This changeset implements subselect objects that can be used in a join like a virtual table or passed to `Expression.belongs()`. Subselect objects can be created using `Set.nested_select()` which accepts the same arguments as regular `select()`. One additional argument, `correlated`, determines whether the subselect may reference tables from parent select (defaults to `True` which means referencing is permitted). Example usage:

    total = table1.field2.sum().with_alias('total')
    sub1 = db(table1.field2 != None).nested_select(table1.field1, total, groupby=table1.field1)
    sub1 = sub1.with_alias('tmp')
    result1 = db(table2.field1 == sub1.field1).select(table2.ALL, sub1.total)

    total = table1.field2.sum().with_alias('total')
    sub2 = db(table1.field2 != None).nested_select(total, groupby=table1.field1)
    result2 = db(table2.field2.belongs(sub2)).select(table2.ALL)

----

I'm opening this pull request early for design discussion. Unit tests for Python 2.7, 3.4 & 3.5 and SQLite, MySQL and PostgreSQL are passing on my machine but there's still work to do. Comments and suggestions are appreciated.
## Design changes
- Adapters and dialects must no longer resolve table names over and over through `DAL.__getitem__()`. They must pass Table/Select objects around.
- New methods for class `Table`:
  - `query_name()` - Full table name for query FROM clause, including possible alias.
  - `query_alias()` - The appropriate string that identifies the table throughout a query (e.g. for building fully qualified field names).
- `Rows.__init__()` now takes additional argument - list of `Field`/`Expression` objects corresponding to columns
- `BaseAdapter._select()` now returns a `Select` object instead of string
- `BaseAdapter.tables()` now returns a dict mapping table names to table objects instead of name list.
- `BaseAdapter.common_filter()` now accepts table objects as well as table names.
- many other `BaseAdapter` internal methods now take or return table objects instead of table names.

### DIFF
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -126,14 +126,14 @@ class BaseAdapter(with_metaclass(AdapterMeta, ConnectionPool)):
         return tables
 
     def get_table(self, *queries):
-        tablenames = self.tables(*queries)
-        if len(tablenames) == 1:
-            return tablenames.popitem()[0]
-        elif len(tablenames) < 1:
+        tablemap = self.tables(*queries)
+        if len(tablemap) == 1:
+            return tablemap.popitem()[1]
+        elif len(tablemap) < 1:
             raise RuntimeError("No table selected")
         else:
             raise RuntimeError(
-                "Too many tables selected (%s)" % str(tablenames))
+                "Too many tables selected (%s)" % str(list(tablemap)))
 
     def common_filter(self, query, tablist):
         tenant_fieldname = self.db._request_tenant
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -509,25 +509,26 @@ class SQLAdapter(BaseAdapter):
         (rid._table, rid._record) = (table, None)
         return rid
 
-    def _update(self, tablename, query, fields):
+    def _update(self, table, query, fields):
         sql_q = ''
+        tablename = table.sqlsafe
+        query_env = dict(current_scope=[tablename])
         if query:
             if use_common_filters(query):
-                query = self.common_filter(query, [tablename])
-            sql_q = self.expand(query)
+                query = self.common_filter(query, [table])
+            sql_q = self.expand(query, query_env=query_env)
         sql_v = ','.join([
-            '%s=%s' % (field.sqlsafe_name, self.expand(value, field.type))
+            '%s=%s' % (field.sqlsafe_name,
+                self.expand(value, field.type, query_env=query_env))
             for (field, value) in fields])
-        tablename = self.db[tablename].sqlsafe
         return self.dialect.update(tablename, sql_v, sql_q)
 
-    def update(self, tablename, query, fields):
-        sql = self._update(tablename, query, fields)
+    def update(self, table, query, fields):
+        sql = self._update(table, query, fields)
         try:
             self.execute(sql)
         except:
             e = sys.exc_info()[1]
-            table = self.db[tablename]
             if hasattr(table, '_on_update_error'):
                 return table._on_update_error(table, query, fields, e)
             raise e
@@ -536,17 +537,18 @@ class SQLAdapter(BaseAdapter):
         except:
             return None
 
-    def _delete(self, tablename, query):
+    def _delete(self, table, query):
         sql_q = ''
+        tablename = table.sqlsafe
+        query_env = dict(current_scope=[tablename])
         if query:
             if use_common_filters(query):
-                query = self.common_filter(query, [tablename])
-            sql_q = self.expand(query)
-        tablename = self.db[tablename].sqlsafe
+                query = self.common_filter(query, [table])
+            sql_q = self.expand(query, query_env=query_env)
         return self.dialect.delete(tablename, sql_q)
 
-    def delete(self, tablename, query):
-        sql = self._delete(tablename, query)
+    def delete(self, table, query):
+        sql = self._delete(table, query)
         self.execute(sql)
         try:
             return self.cursor.rowcount
@@ -772,18 +774,19 @@ class SQLAdapter(BaseAdapter):
     def _count(self, query, distinct=None):
         tablemap = self.tables(query)
         tablenames = list(tablemap)
-        tables = tablemap.values()
+        tables = list(tablemap.values())
+        query_env = dict(current_scope=tablenames)
         sql_q = ''
         if query:
             if use_common_filters(query):
                 query = self.common_filter(query, tables)
-            sql_q = self.expand(query)
-        sql_t = ','.join(self.table_alias(t, tablenames) for t in tables)
+            sql_q = self.expand(query, query_env=query_env)
+        sql_t = ','.join(self.table_alias(t, []) for t in tables)
         sql_fields = '*'
         if distinct:
             if isinstance(distinct, (list, tuple)):
                 distinct = xorify(distinct)
-            sql_fields = self.expand(distinct)
+            sql_fields = self.expand(distinct, query_env=query_env)
         return self.dialect.select(
             self.dialect.count(sql_fields, distinct), sql_t, sql_q
         )
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -10,7 +10,7 @@ from ..connection import ConnectionPool
 from ..exceptions import NotOnNOSQLError
 from ..helpers.classes import Reference, ExecutionHandler, SQLCustomType, \
     SQLALL, NullDriver
-from ..helpers.methods import use_common_filters, xorify
+from ..helpers.methods import use_common_filters, xorify, merge_tablemaps
 from ..helpers.regex import REGEX_SELECT_AS_PARSER, REGEX_TABLE_DOT_FIELD
 from ..migrator import Migrator
 from ..objects import Table, Field, Expression, Query, Rows, IterRows, \
@@ -114,15 +114,15 @@ class BaseAdapter(with_metaclass(AdapterMeta, ConnectionPool)):
 
     def tables(self, *queries):
         tables = dict()
-        # FIXME: check for name conflicts
         for query in queries:
             if isinstance(query, Field):
-                tables[query.tablename] = query.table
+                key = query.tablename
+                if tables.get(key, query.table) is not query.table:
+                    raise ValueError('Name conflict in table list: %s' % key)
+                tables[key] = query.table
             elif isinstance(query, (Expression, Query)):
-                if query.first is not None:
-                    tables.update(self.tables(query.first))
-                if query.second is not None:
-                    tables.update(self.tables(query.second))
+                tmp = [x for x in (query.first, query.second) if x is not None]
+                tables = merge_tablemaps(tables, self.tables(*tmp))
         return tables
 
     def get_table(self, *queries):
@@ -565,15 +565,17 @@ class SQLAdapter(BaseAdapter):
         for item in param:
             if isinstance(item, Expression):
                 item = item.first
-            tablemap[item._tablename] = item
+            key = item._tablename
+            if tablemap.get(key, item) is not item:
+                raise ValueError('Name conflict in table list: %s' % key)
+            tablemap[key] = item
         join_tables = [
             t._tablename for t in param if not isinstance(t, Expression)
         ]
         join_on = [t for t in param if isinstance(t, Expression)]
         tables_to_merge = {}
         for t in join_on:
-            # FIXME: check for name conflicts
-            tables_to_merge.update(self.tables(t))
+            tables_to_merge = merge_tablemaps(tables_to_merge, self.tables(t))
         join_on_tables = [t.first._tablename for t in join_on]
         for t in join_on_tables:
             if t in tables_to_merge:
@@ -603,8 +605,7 @@ class SQLAdapter(BaseAdapter):
             query = self.expand(query)
         #: auto-adjust tables
         for field in fields:
-            # FIXME: check for name conflicts
-            tablemap.update(self.tables(field))
+            tablemap = merge_tablemaps(tablemap, self.tables(field))
         if len(tablemap) < 1:
             raise SyntaxError('Set: no tables selected')
         #: remove outer scoped tables if needed
@@ -627,15 +628,13 @@ class SQLAdapter(BaseAdapter):
                 ijoin_tables, ijoin_on, itables_to_merge, ijoin_on_tables,
                 iimportant_tablenames, iexcluded, itablemap
             ) = self._build_joins_for_select(tablemap, join)
-            # FIXME: check for name conflicts
-            tablemap.update(itablemap)
+            tablemap = merge_tablemaps(tablemap, itablemap)
         if left:
             (
                 join_tables, join_on, tables_to_merge, join_on_tables,
                 important_tablenames, excluded, jtablemap
             ) = self._build_joins_for_select(tablemap, left)
-            # FIXME: check for name conflicts
-            tablemap.update(jtablemap)
+            tablemap = merge_tablemaps(tablemap, jtablemap)
         current_scope = outer_scoped + list(tablemap)
         table_alias = lambda name: tablemap[name].query_name(current_scope)[0]
         if join and not left:
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -595,10 +595,9 @@ class SQLAdapter(BaseAdapter):
                       cache=None, cacheable=None, processor=None):
         #: parse tablemap
         tablemap = self.tables(query)
-        tablenames_for_common_filters = tablemap
         #: apply common filters if needed
         if use_common_filters(query):
-            query = self.common_filter(query, tablenames_for_common_filters)
+            query = self.common_filter(query, list(tablemap.values()))
         #: expand query if needed
         if query:
             query = self.expand(query)
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -154,7 +154,8 @@ class BaseAdapter(with_metaclass(AdapterMeta, ConnectionPool)):
                         query = query & newquery
         return query
 
-    def _expand(self, expression, field_type=None, colnames=False):
+    def _expand(self, expression, field_type=None, colnames=False,
+        query_env={}):
         return str(expression)
 
     def expand_all(self, fields, tabledict):
@@ -419,7 +420,8 @@ class SQLAdapter(BaseAdapter):
             handler.after_execute(command)
         return rv
 
-    def _expand(self, expression, field_type=None, colnames=False):
+    def _expand(self, expression, field_type=None, colnames=False,
+        query_env={}):
         if isinstance(expression, Field):
             et = expression.table
             if not colnames:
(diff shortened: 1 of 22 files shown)


pydal/adapters/base.py
```diff
@@ -618,49 +624,60 @@ def _select_wcols(self, query, fields, left=False, join=False,
         #: build joins (inner, left outer) and table names
         if join:
             (
+                # FIXME? ijoin_tables is never used
                 ijoin_tables, ijoin_on, itables_to_merge, ijoin_on_tables,
```
nextghost: Possible bug: ijoin_tables is never used.

pydal/adapters/base.py
```diff
@@ -618,49 +624,60 @@ def _select_wcols(self, query, fields, left=False, join=False,
...
+            tokens += [self.dialect.cross_join(table_alias(t))
+                    for t in cross_joins[1:]]
+            # FIXME: WTF? This is not correct syntax at least on PostgreSQL
             if join_tables:
-                sql_t += ' %s' % \
-                    self.dialect.left_join(','.join([t for t in join_tables]))
-            for t in join_on:
-                sql_t += ' %s' % self.dialect.left_join(t)
+                tokens.append(self.dialect.left_join(','.join([table_alias(t)
+                        for t in join_tables])))
```
nextghost: Query syntax like `LEFT JOIN tab1,tab2` without any `ON` clause doesn't work in PostgreSQL and possibly other databases. If some databases do support this syntax, it should be moved to corresponding dialects. Don't forget to update the other copy of this code below as well.

pydal/objects.py
```diff
@@ -2122,7 +2269,7 @@ def validate_and_update(self, **update_fields):
         return response
 
     def delete_uploaded_files(self, upload_fields=None):
-        table = self.db[self.db._adapter.tables(self.query)[0]]
+        table = self.db._adapter.tables(self.query).popitem()[1]
```
nextghost: The new code is functionally equivalent to the original but neither version will work properly if there's more than 1 table in the query.

niphlod:
boys I so don't like _select() not returning a string. People may have been grow fond of that.

On another entire matter, I don't feel writing that much of code over a raw t-sql when in need of such complications...the proposed piece of pydal code to write is maybe longer than the query itself, and it's not readable at all.
Also, writing adapters for all backends could be a nightmare. I feel that a relevant part of the problems you can solve with correlated subselects would be better served by window functions.

gi0baro:
@niphlod on the other hand I find the idea quite interesting, but I'm thinking about it in the matter of how pydal builds sql strings.
I think would be quite powerful to change the way the dialects builds sql strings, 'cause using custom classes and objects (that could implement the `__str__` method and behave like strings with `__repr__` so users wouldn't notice the difference) would allow us to really customize the sql generation. Moreover this opens the chance to implement caching on sql generation, and to simplify the flow of quoting-unquoting strings when talking to the driver (= performance improvements).

I think we should really move the discussion on a different "floor" here, but is definitely another story from this PR and the idea behind it. @mdipierro what do you think?

mdipierro:
_select MUST return a string. it is in the specs.

select() makes a query and _select() gives me the corresponding SQL. Same for update and _update, insert and _insert, count and _count. 

That is also needed for subselects like db(db.table.field.belongs(db(subquery)._select(db.table.field)).select()

We should not change it. It is ok to return an object that extends str.

nextghost:
If I understand correctly how Python strings work, making the `Select` class an extension of `str` would require building the full SQL subquery string immediately in `Select.__new__()` and never changing it afterwards. I don't think that's a good idea and preparing subselects that reference tables from parent query would be more complex than it needs to be. Expanding a `Select` object in the `belongs()` expression is already implemented in `SQLDialect.belongs()`. Only Mongo and Google adapters reimplement that and they don't seem to support nested selects there anyway. But if you insist on `_select()` returning a string, I'll just revert the one-line change there and move it to `nested_select()` instead.

mdipierro:
Yes. I would like that better. No need to break the current functionality since we can have a nested_select that is more clear anyway.

nextghost:
It appears that nobody actually tested Rows.render(). Unit tests call it but don't check the output. If you called it as `result.render(fields=[tab.f1, tab.f2, tab.f3])`, those three fields would be rendered three times each. Fixed in the refactored version.

gi0baro:
@mdipierro I'm for changing the behaviour: I think it's quite useless to stay on strings if we want to alter queries *on the go*, because it's just loosing time in building a string that will be rebuilt again and again. Having a `SQLString` object that **behaves** like a string, and that can be cached and takes care of quoting sounds quite promising regarding performance and customization.

@nextghost maybe before you continue with subselects we can design something more generic for the way strings are built in pyDAL, so that you can benefit from the new structure to implement subselects?

@niphlod opinions?

nextghost:
GAE adapter should be working again, at least according to unit tests. IMAP, Mongo and CouchDB adapters don't seem to support joins anyway so I'll skip refactoring them. However, I highly recommend refactoring the rest of `self.db[tablename]` anti-pattern later. It can break if you get too creative with table aliasing.

Also, `BaseAdapter` should probably have its own `nested_select()` method so that NoSQL adapters can throw `NotOnNOSQLError` from it.

@gi0baro I'm almost done with adapter refactoring so there's no point in waiting for another big redesign. Just 2 or 3 more commits and I can wrap it up by writing unit tests. But here are some tips you might find useful:

- I recommend using different subclasses of `Expression` for each operation rather than storing a dialect method in the `Expression` object. E.g. the `AndExpression` should call `Dialect._and()` directly. Then you can easily test what kind of expression you're looking at using `isinstance()`.
- The new interface should have preliminary support for using placeholders in the finished query. For example the query expansion method should return a tuple `('SQL string', [placeholder values])`. If you look at `Select.query_name()` in this PR, it already works like that.
- The return value of `Table.on()` should not be an instance of `Expression`. You should probably keep it only for backward compatibility and instead implement `Join` objects that allow arbitrary tree of joins. The current interface for example doesn't allow you to do this: `SELECT ... FROM a LEFT JOIN (b INNER JOIN c ON ...) ON ...`.

niphlod:
@gi0baro : on top of the notes on https://github.com/web2py/pydal/issues/411, I don't think that caching the generated query will solve THAT amount of performance to be noticeable. 
Finalizing on the "_select() returning a string", I'd say that if 
q = db(db.table.id>0)._select()
db.executesql(q[:-1])
works with what you want to make _select() do, we should be fine.

If we want to really save time, it should be saved in avoiding expanding arguments by ourselves, as the 99% of the drivers are mature enough to do it by themselves (and query plans for prepared statements are cached on the backend, that's one real speedup).  It'll eventually close also the requirement for using binaries instead of base64 encoding them (and all the memory issues in heavy-processing pieces of code).

nextghost:
Adapter refactoring should be complete. There are some FIXMEs in the changeset that you should review and fix later (fixing the one about name conflicts will probably break existing third-party code). The new code for passing parent scope tables to subqueries through expression trees isn't pretty but since the whole thing is up for a major redesign anyway, it'll do for now.

Summary of internal changes: Many adapter methods that originally accepted/returned table names now accept/return table objects. However, some code still uses the `self.db[tablename]` anti-pattern when it should pass the table object around directly.

In addition to the usual `select()` parameters, the `nested_select()` method also accepts a new parameter `correlated` which defaults to `True`. If set to `False`, the subquery will never reference any tables from the parent scope. Otherwise tables will be resolved in this order: 1) tables appearing in `left` or `join` arguments; 2) tables from parent scope; 3) tables appearing in `query` or `fields` arguments.

Note that the FROM clause may become empty as a result. That is perfectly valid SQL syntax but not currently supported by PyDAL. If you want to have the same table in the FROM clause of both the parent query and the subquery while allowing references to parent scope, you'll have to use unique aliases.

All that's left for me to do is to write unit tests. If you have any further suggestions for this PR, now's your last chance.

stephenrauch:
So I am late to this conversation, but I implemented a significant portion of the mongo driver.  I was looking at the changes covered here, and it seems the majority of them in the mongo code, are the passing of the parameter 'query_env='.  But I failed to find that value used anywhere.  Sorry if this was already covered someplace, and I missed it, but what is the purpose of this data structure?  Also several of the functions have their parameter list as 'query_env={}'.  It is generally bad form to have a mutable as a default value in a parameter list.

Thanks...

nextghost:
@stephenrauch  NoSQL databases don't support subselects so all the changes in Mongo code are only interface updates. `query_env` is currently used by SQL dialects to correctly expand subselects which reference tables from parent query. The contents are dialect-specific and you can use it as you see fit. See `SQLAdapter._select_wcols()` for what is passed in it and `SQLDialect.on()` or `SQLDialect.belongs()` for how the contents are currently used. Using `query_env` to [pass data back](https://en.wikipedia.org/wiki/Visitor_pattern) from the expression tree would also be a valid use case.

Using e.g. `None` as the default value for `query_env` would require either additional checks whether the parameter can be used or much larger refactoring of adapter code. @gi0baro already proposed a redesign of query string construction so you can find a better solution later.

stephenrauch:
@nextghost 

Thanks for the reply.

Aa an FYI, Mongo is in fact capable of doing the equivalent of a subselect using its aggregation pipeline.  It may not be worth the bother for pydal to support that, but it is possible.

The 'query_env' parameter appears to start at the top of the tree through _expand() via expand().  The expand method is only called from SQLAdapter or from specific adapters.  So it does not appear to me that all of Mongo's expand needs this interface change at all.  And to my eye it considerably clutters the code. The string 'query_env' now appears 174 times in a file which only contains 461 lines of code.  And yet 'query_env' is not used once in all of that.

At the larger level in this PR, the string 'query_env' appears 753 times, and yet is only used a very few times.  In an effort to avoid a bunch of boilerplate, the mongo adapter uses the decorators 'needs_aggregation_pipeline()' and 'check_fields_for_cmp()', as well as an attribute '_parse_data' attached to Expression, Query and their children, to track some meta data without needing to modify (clutter) dozens of parse functions that have no need to understand the larger context. A similar mechanism might allow this PR to perform the same function as the current suggested code, without so much of an impact on non-associated source code.

Thanks again.

@gi0baro @niphlod @mdipierro 

So to summarize the above:

I would strongly suggest that the extensive changes to Mongo driver with the addition of 'query_env' are not warranted.  

I would also (weakly) suggest that 'query_env' could be made an attribute of Expression and Query to allow it to be accessed in the depths of the parse tree without needing to be passed around explicitly to so many places where it does not actually matter.

nextghost:
@stephenrauch If you pass `query_env` as an attribute, you may end up with extremely complex side-effects. I assume you want to do something like this:

    class SQLAdapter:
        def _expand(self, expression, field_type=None, colnames=False):
            if isinstance(expression, (Expression, Query)):
                first = expression.first
                second = expression.second
                first.query_env = second.query_env = expression.query_env
                op = expression.op
                rv = op(first, second)
                ...

Remember that Expression objects can be reused and the whole structure may be a general acyclic graph, not a tree. If `second` is reused somewhere inside `first` and one of the operations along the way replaces the `query_env` object with something else, calling `expand(first)` will trash `second.query_env` before control flow reaches `expand(second)`. Or vice versa.

stephenrauch:
@nextghost 

I was intending to make query_env an attribute of second that references the query_env that is actually an attribute of the Query or expression, just as your example states,  But, my assumption was that the actual query_env object (dictionary), would not be replaced, but extended or modified... Also necessarily, it would need to be assigned to the Query or Expression at the appropriate time in the graph.  However, if second's (in your example) query_env does need to be replaced some place in the graph, it would simply need to be restored on the way back out.

So...  the life span of query_env passed as a parameter to the function and as attached to 'second', from my perspective, can be made identical.  Thus any side effects will be the same in both cases.  By decorating the functions that need to do particular operations on query_env, it is straightforward to attach/detach query_env from the parameters as needed, and then this context code is largely isolated to the decorator and thus minimizes the boilerplate code of constantly passing query_env.

Thanks again.

PS.  I don't understand how this is a graph but not a tree, since there is a root, and parsing is done recursively.  I would be grateful for some education on that topic...

nextghost:
@stephenrauch Even if we implement `query_env` as an attribute correctly now, there's still the danger that somebody will forget about the invariants and use it as a persistent attribute later. Passing function arguments through a transient object attribute that gets assigned behind the scenes is quite counter-intuitive.

Tree is a graph where there is exactly one (unoriented) path between any pair of vertices. Directed acyclic graph is an oriented graph that contains no oriented cycles but there may exist multiple distinct oriented paths between a pair of vertices (all paths going in the same direction). See [Wikipedia](https://en.wikipedia.org/wiki/Directed_acyclic_graph) for details and examples.

In other words, a structure of `Expression` objects may not be a tree because two distinct paths from the root can converge at the same subexpression.

gi0baro:
Anyway @nextghost I'm for releasing a 16.9 release with bugfixes and some optimizations before merging this. So practically this is targeted for 16.10

@mdipierro do you agree?

niphlod:
/me agrees . unrelated: @gi0baro : we need to address some intricacies found when executing pydal's test under web2py: there are some quacky tests in pydal's test/sql.py and - ideally - I'd like to get them fixed before releasing 16.09

nextghost:
@niphlod If you're going to dig into unit tests, I'll open another PR for custom unittest subclass that takes care of the database connect and cleanup boiler plate. When I have to manually drop tables from PostgreSQL and delete the SQLite database after every failed unit test, it's a real pain in the ass.

niphlod:
everything that "smarts" is always welcome

mdipierro:
@gi0baro I agree. You merge this.

nextghost:
I have to take care of some other urgent work so it'll be a while before I can finish the unit tests. Meanwhile, what do you all think about `query_env`? Leave it as is or pass the data some other way?

gi0baro:
@nextghost take your time. I'm quite busy too during this month, will update you with feedbacks as soon as I can review your changes completely.

niphlod:
this should also fix https://github.com/web2py/web2py/issues/696

nextghost:
@niphlod Sorry, no. The bug is very much still present here. This pull request is already huge and I will not add another 200 lines of code just to fix something I didn't break myself.

We can discuss the solution in issue #411. I'll open a new pull request once this gets merged.

mdipierro:
what's the status here?

niphlod:
@mdipierro : everything scheduled for the next or the following release (meaning hopefully for december)

nextghost:
@mdipierro I still haven't had time to finish the unit tests. If things go well, I might get it done next week.

nextghost:
Aaaand, done. If you don't have any further comments or suggestions for this pull request, you can merge it.

gi0baro:
@nextghost @mdipierro will review this carefully in the next days.

pydal/adapters/base.py
```diff
@@ -579,37 +593,28 @@ def _build_joins_for_select(self, tablenames, param):
...
+        tablemap = merge_tablemaps(tablemap, self.tables(*fields))
         #: remove outer scoped tables if needed
-        if outer_scoped:
-            tablenames = [
-                name for name in tablenames if name not in outer_scoped]
-        #: prepare columns and expand fields
-        colnames = list(map(self._colexpand, fields))
-        sql_fields = ', '.join(map(self._geoexpand, fields))
+        for item in outer_scoped:
+            # FIXME: check for name conflicts
```
gi0baro: @nextghost what about this? Should be fixed?

pydal/adapters/base.py
```diff
@@ -618,55 +623,71 @@ def _select_wcols(self, query, fields, left=False, join=False,
...
+            sql_t = ' '.join(tokens)
         elif not join and left:
             cross_joins = excluded + list(tables_to_merge)
-            sql_t = '%s' % self.table_alias(cross_joins[0])
-            for t in cross_joins[1:]:
-                sql_t += ' %s' % self.dialect.cross_join(self.table_alias(t))
+            tokens = [table_alias(cross_joins[0])]
+            tokens += [self.dialect.cross_join(table_alias(t), query_env)
+                    for t in cross_joins[1:]]
+            # FIXME: WTF? This is not correct syntax at least on PostgreSQL
```
gi0baro: @nextghost can you explain this better? Which is the correct syntax?

pydal/adapters/base.py
```diff
@@ -618,55 +623,71 @@ def _select_wcols(self, query, fields, left=False, join=False,
...
-            sql_t = '%s' % self.table_alias(tables_not_in_joinon[0])
-            for t in tables_not_in_joinon[1:]:
-                sql_t += ' %s' % self.dialect.cross_join(self.table_alias(t))
-            for t in ijoin_on:
-                sql_t += ' %s' % self.dialect.join(t)
+            tokens = [table_alias(tables_not_in_joinon[0])]
+            tokens += [self.dialect.cross_join(table_alias(t), query_env)
+                    for t in tables_not_in_joinon[1:]]
+            tokens += [self.dialect.join(t, query_env) for t in ijoin_on]
+            # FIXME: WTF? This is not correct syntax at least on PostgreSQL
```
gi0baro: @nextghost same as the other comment

pydal/objects.py
```diff
@@ -1039,6 +1053,152 @@ def drop_index(self, name):
...
+            colnames, sql = adapter._select_wcols(self._query, self._qfields,
+                **attributes)
+            # Do not cache when the query may depend on external tables
+            if not outer_scoped:
+                self._colnames_cache, self._sql_cache = colnames, sql
+        else:
+            colnames, sql = self._colnames_cache, self._sql_cache
+        if with_alias and self._tablename is not None:
+            alias = self._db._adapter.dialect.quote(self._tablename)
+            if 'Oracle' in str(type(self._db._adapter)):
```
gi0baro: @nextghost you should use dialect's `_as` method instead of checking the adapter type

gi0baro:
@nextghost can you also please update the description of this PR, also adding a quick example of the usage?

pydal/adapters/base.py
nextghost: Fixing it would break backward compatibility with third-party code. `outer_scoped` is currently a list of internal table names. It would have to be changed into a dict of table objects, otherwise there is no way to check for conflicts (same name referencing different objects).

pydal/adapters/base.py
nextghost: `join_tables` gets expanded below the comment into something like `LEFT JOIN tab1, tab2, tab3, ...`. MySQL and PostgreSQL both raise syntax error on that but SQLite seems to accept it. At the very least, the expansion should be handled by the dialect, not `SQLAdapter._select_wcols()`.

pydal/objects.py
nextghost: This bit was copied from `Table.__str__()`. Should I fix both?

pydal/adapters/base.py
gi0baro: @nextghost ok, will inspect that apart from this PR

pydal/objects.py
gi0baro: @nextghost would be awesome :)

nextghost:
@gi0baro I've fixed table/subselect alias expansion as you requested and added another minor bugfix. I've also updated PR description and added usage example. (Should I remove the old description as well?)

But before you merge it, you should also take another look at the two pieces of code which I've highlighted right below the PR description.

gi0baro:
Thank you @nextghost! The description looks good to me right now.
I will update this later during the week regarding your two comments and some benchmarks.
I feel quite sorry about the *slowness* of this but it's quite a large PR you know? :)

gi0baro:
@nextghost ok, the performance seems quite the same from my benchmarks.
Regarding your comments, I would like to open up two new issues for those (as for the wrong syntax about the left joins) and discuss about them separately from this.

If you don't have any other considerations, I will merge this.

nextghost:
I have nothing more to add, you can merge this.

PR #415, opened 2016-09-09
New test helper class that handles connection boilerplate

nextghost:
This patch introduces a new unittest class `DALtest` which handles connection boilerplate and cleans up automatically after each test run, whether it ended successfully or crashed. It implements 2 methods:

- `connect()` - create new `DAL` instance and register it for cleanup. All arguments are passed to `DAL` constructor. `uri` defaults to `DEFAULT_URI`. `check_reserved` defaults to `['all']`. Do ***not*** call `close()` on the returned object manually!
- `tearDown()` - drop all tables and close all connections created with `connect()`. It is safe to call this manually in the middle of a test run in order to purge current connections.

I've refactored `tests/sql.py` to use the new class. There are a few cases where opening and closing of connections was too complex for a quick & simple replacement so I didn't touch those. Refactoring of other test scripts was left as an exercise.

### DIFF
--- a/tests/_helpers.py
+++ b/tests/_helpers.py
@@ -1,0 +1,25 @@
+from ._compat import unittest
+from ._adapt import DEFAULT_URI
+from pydal import DAL
+
+class DALtest(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super(DALtest, self).__init__(*args, **kwargs)
+        self._connections = []
+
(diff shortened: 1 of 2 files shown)


mdipierro:
Thanks for this commit. I am ok with this change but before I merge it, anybody opposes or thinks DALTest belongs somewhere else? Speak now.

PR #420, opened 2016-09-19
fix for opening files in py3

niphlod:
fixes also https://github.com/web2py/web2py/issues/1451

### DIFF
--- a/pydal/contrib/portalocker.py
+++ b/pydal/contrib/portalocker.py
@@ -58,7 +58,7 @@ Web2py Changes
 import sys
 import logging
 
-PY2 = sys.version_info[2] == 2
+PY2 = sys.version_info[0] == 2
 
 logger = logging.getLogger("pydal")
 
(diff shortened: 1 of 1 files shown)


gi0baro:
@niphlod ready for merge?

niphlod:
Yep

Il 20 set 2016 11:28, "Giovanni Barillari" <notifications@github.com> ha
scritto:

pydal/contrib/portalocker.py
```diff
@@ -55,9 +55,13 @@
      solution
 
 """
-
+import sys
 import logging
-logger = logging.getLogger("web2py")
+
+PY2 = sys.version_info[2] == 2
```
ilvalle: It is ```sys.version_info[0] == 2```

niphlod:
@ilvalle : fixed now. Dunno if tests should be added here or on web2py

ilvalle:
portalocker is used on both pydal and web2py. Placing unittest in one or another is the same, perhaps in web2py is better because language files can be used as input for tests.

PR #429, opened 2016-11-21
fix issue where Set.delete_uploaded_files removes files it should not…

leonelcamara:
… during an update

The change I made differentiates between updates and deletes by taking advantage of upload_fields always being None in the case of deletes. If it is an update it only deletes the file if the field with it was updated and the new value is different.

See:
https://groups.google.com/forum/#!topic/web2py-developers/gwEr1cHQNAY

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -2141,24 +2141,23 @@ class Set(Serializable):
                 oldname = record.get(fieldname, None)
                 if not oldname:
                     continue
-                if upload_fields and fieldname in upload_fields and \
-                   oldname == upload_fields[fieldname]:
-                    continue
-                if field.custom_delete:
-                    field.custom_delete(oldname)
-                else:
-                    uploadfolder = field.uploadfolder
-                    if not uploadfolder:
-                        uploadfolder = pjoin(
-                            self.db._adapter.folder, '..', 'uploads')
-                    if field.uploadseparate:
-                        items = oldname.split('.')
-                        uploadfolder = pjoin(
-                            uploadfolder, "%s.%s" %
-                            (items[0], items[1]), items[2][:2])
-                    oldpath = pjoin(uploadfolder, oldname)
-                    if exists(oldpath):
-                        os.unlink(oldpath)
+                if upload_fields is None or (fieldname in upload_fields and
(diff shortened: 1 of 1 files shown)


leonelcamara:
I'm still not sure this is the complete fix as it is possible that an update causes compute fields to change and they will not appear in upload_fields, causing the file to not be deleted.

pydal/objects.py
```diff
@@ -2141,24 +2148,23 @@ def delete_uploaded_files(self, upload_fields=None):
                 oldname = record.get(fieldname, None)
                 if not oldname:
                     continue
-                if upload_fields and fieldname in upload_fields and \
-                   oldname == upload_fields[fieldname]:
-                    continue
```
gi0baro: @leonelcamara why changing this whole block of code to avoid a `continue`?

gi0baro:
@leonelcamara @mdipierro @niphlod can I start a discussion about computed fields before merging this?

I would like to change the whole logic of how computed fields are computed during inserts/updates 'cause for example we still can't use them in callbacks (see https://github.com/gi0baro/weppy/issues/151 for example). May I produce a bigger PR including this fix? What do you think?

niphlod:
The more streamlined the better

gi0baro:
@niphlod what do you mean by *streamlined*? :)

leonelcamara:
> I would like to change the whole logic of how computed fields are computed during inserts/updates 'cause for example we still can't use them in callbacks (see gi0baro/weppy#151 for example). May I produce a bigger PR including this fix? What do you think?

@gi0baro sure, I think we need to discuss computed fields, because although this PR fixes a problem it is not efficient. Fields with a compute should probably have a required_fields for instance.

niphlod:
Streamlined as in "we know current has drawbacks and we'll think about fix
it once and for all" :-P

mdipierro:
@leonelcamara @gi0baro this PR now has conflicts with the one that was merged. Can you please help resolve those?

gi0baro:
@leonelcamara @mdipierro I close this since is already fixed by #431

mdipierro:
thanks!

PR #431, opened 2016-12-02
Store values for callbacks on insert/update/delete operations

gi0baro:
Ok everybody, this replace the current behaviour of `Table._listify`:
- all the values computed before callbacks are passed to callbacks
- removed duplicated logic for default fields: removed `_defaults` method
- all the fields that have callable values are computed in the same flow (defaults, updates, computations)

Rationale:
- Fixing computed fields flow (see #429 )
- Allow to use computed fields inside callbacks (see https://github.com/gi0baro/weppy/issues/151 and #429)
- Have a unique flow for callable values

Breaking change:
- If you accessed a callable default or update value in callbacks you had to invoke that within the callback to get the real value. Now is just the real value. 

Even if this is a breaking change, I consider the old behaviour *wrong*, since, for example, if you had a *created\_at* field with `default=lambda: datetime.utcnow()`, and in an *after_insert* callback you wanted to access that value in the fields, you simply couldn't, since the real value inserted calling the method is not stored in the flow.

I would like to hear @mdipierro, @niphlod, @leonelcamara on this.

This changed also the flow regarding `Table._attempt_upload`, since now is a callback method in the `before_insert` and `before_update`. So while the old flow was:
`_attempt_upload`, `callable_values`, `before_callback`, `operation`

the new one is:
`callable_values`, `before_callback`, `operation`

### DIFF
--- a/pydal/helpers/classes.py
+++ b/pydal/helpers/classes.py
@@ -5,13 +5,26 @@ import struct
 import time
 import traceback
 
-from .._compat import PY2, exists, copyreg, integer_types, implements_bool, \
-    iterkeys, itervalues, iteritems
+from .._compat import (
+    PY2, exists, copyreg, implements_bool, iterkeys, itervalues, iteritems,
+    long
+)
 from .._globals import THREAD_LOCAL
 from .serializers import serializers
 
 
-long = integer_types[-1]
+class cachedprop(object):
+    #: a read-only @property that is only evaluated once.
+    def __init__(self, fget, doc=None):
+        self.fget = fget
+        self.__doc__ = doc or fget.__doc__
+        self.__name__ = fget.__name__
+
+    def __get__(self, obj, cls):
+        if obj is None:
+            return self
+        obj.__dict__[self.__name__] = result = self.fget(obj)
+        return result
 
 
 @implements_bool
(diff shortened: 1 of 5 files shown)


gi0baro:
@mdipierro @niphlod @leonelcamara ..comments?

niphlod:
bummer, I reeeeaally don't have time to scour the code. IMHO both the rationale and the breaking change is fine. 
I don't really have the time to assert where _attempt_upload should be. 
Breaking something that theoretically should be atomic (the CRUD operation in the db AND the CRUD operation in the file) in all code-paths (stored to disk, stored in a virtual FS, converted to base64 and injected in another column) is something that, no matter how, is really error-prone given the flexibility uploads have (don't even get me started on storing the upload in a real binary field, which is a huuuuuge miss on dal/pydal since its conception).
The only thing I'm midly sure about is that NOT needs to happen after the _before_callback (or, in reverse words, it needs to happen either before or in _before_callback)

gi0baro:
@niphlod I've just moved the `attempt_upload` into the before callbacks.

@mdipierro @leonelcamara I consider this ready to be merged. Still waiting for your comments.

gi0baro:
@mdipierro I will merge this by Thu 22/12 unless you or @leonelcamara have any comments.

mdipierro:
LGTM

Issue #430, opened 2016-11-21
Possible typo in objects.py in Field.clone()


patrakov:
Regarding self-reference handling in Field.clone(), I find it very strange that the following subexpression exists in the code:

    if ... field.type == 'reference %s'+field._tablename ...:

Comparing the type to "reference %sfoo" looks wrong to me. I think you meant:

    if ... field.type == 'reference %s' % field._tablename ...:

Issue #57, opened 2015-02-11
Having a problem with `with_alias`


rcalsaverini:
Hi there. I'm trying to use pydal with an existing set of tables on my database and I'm hitting a wall here about how to deal with two rows of the same table appearing in the same query.

First a working example. Imagine I have two tables representing a directed graph: 

```python
from pydal import * 

db = DAL('mysql+mysqldb://someuser:somepassword@somehost:3306/graph')

db.define_table(
    'node',
    Field('name', type="string"),
    rname="someschema.some_rname_for_node"   
)

db.define_table(
    'edge',
    Field('edge_id', type="id"),
    Field('node_from', type=db.node),
    Field('node_to', type=db.node),
    rname="someschema.some_rname_for_edges"
)
```

Than I want the simplest query, just to print the names of the nodes that are connected by an edge:
```python
db.node.with_alias('start')
db.node.with_alias('end')
db.edge.with_alias('hop')

hop_join = (db.start.id == db.hop.node_from) & (db.end.id == db.hop.node_to)
print db(hop_join)._select(db.start.name, db.end.name)
```
The obtained query is:

```sql
SELECT 
    start.name, end.name
FROM
    graph.some_rname_for_node AS start,
    graph.some_rname_for_node AS end,
    graph.some_rname_for_edges AS hop
WHERE
    ((start.id = hop.node_from)
        AND (end.id = hop.node_to));
```
Which is exactly what I want. This works perfectly.

But for some reason, when I try this with my real database this doesn't work at all:
```python
db.define_table(
    'position',
    Field('position_id', rname='cargo_id', type='id'),
    Field('title', rname="titulo", type="string"),
    rname='cargo'
)

db.define_table(
    'similar_positions',
    Field('similar_id', rname='cargo_similar_id', type='id'),
    Field('position_from', rname='cargo_id1', type=db.position),
    Field('position_to', rname='cargo_id2', type=db.position),
    rname='cargo_similar'
)

db.position.with_alias("position_from")
db.position.with_alias("position_to")
db.similar_positions.with_alias("hop")

hop = (
    (db.position_from.id == db.hop.position_from) & 
    (db.position_to.id == db.hop.position_to)
)

print db(hop)._select(db.position_from.title, db.position_to.title)
```

The sql generated is:
```sql
SELECT 
    position_from.titulo, position_to.titulo
FROM
    cargo,
    cargo_similar AS hop,
    cargo AS position_from,
    cargo AS position_to
WHERE
[...]

Issue #343, opened 2016-04-18
Error when renaming a field in Postgres table definition


62mkv:
Hi

We are trying to use pyDAL for "code-first" approach (for versioning of database schema).  
First we wrote 

```
db.define_table(
    'settings',
    Field('name', type='string', length=256, required=True),
    Field('admin_password', type='string', length=256, required=True),
    rname='dich.settings',
    migrate='dich.settings'
)
```

And a table was created OK. 

Then we've renamed "admin_password" field to "value": 

```
db.define_table(
    'settings',
    Field('name', type='string', length=256, required=True),
    Field('value', type='string', length=256, required=True),
    rname='dich.settings',
    migrate='dich.settings'
)
```

We've run the script and are getting Posgre error: relation "settings" does not exist.

in the sql.log, we see: 

> ALTER TABLE settings DROP COLUMN admin_password;

Apparently, pyDAL does not provide schema name for such operation. 

Is it our fault or a bug in pyDAL ? Thanks a lot !

gi0baro:
@62mkv This seems a lack of schema name handling by pydal migrator, that uses just the name of the table. Will do some tests to check where we can change this.

niphlod:
@gi0baro : this too should be targeted/inspected before the october-november release (i.e. not the next, but "second next" one)

nextghost:
Will be fixed by PR #433

PR #433, opened 2016-12-19
Table name bugfixes

nextghost:
This changeset fixes update()/delete() on aliased tables and bugs caused by using the wrong table/field name in various queries (especially in migrator).

### Changes in class `Table`
- `Table._ot` has been renamed to `Table._dalname` and it is always set to the original internal table name.
- `Table._rname` now defaults to quoted `Table._dalname` (except when the table is not associated with any database, only then it gets set to `None`).
- Added `Table._raw_rname` which is the unquoted version of `_rname` (for generating extra identifers based on the backend table name, e.g. sequence names).
- `Table.sqlsafe` renamed to `sql_shortref`.
- `Table.sqlsafe_alias` renamed to `sql_fullref`.

Example where they belong in a query: `SELECT * FROM sql_fullref WHERE sql_shortref.id > 0`

### Changes in class `Field`
- `Field.sqlsafe` now raises exception when the field object is not associated with any table.
- Deprecated `Field.sqlsafe_name`, use `Field._rname` instead.
- Added property `Field.longname` which returns field reference string as 'tablename.fieldname' (as oposed to 'table_rname.field_rname' returned by `sqlsafe`).
- `Field.clone()` now returns the new `Field` object without link to any table and with empty `_rname` if it needs to be requoted.
- Added method `Field.bind()` which links the field to a table. When the field is already linked, it raises exception. It also sets empty `Field._rname` to quoted `Field.name`. This method is called internally by table/subselect constructor.
- Added `Field._raw_rname` which is the unquoted version of `_rname`.

### Other changes
- Removed `BaseAdatper.alias()`, the code was move to `Table.with_alias()` where it belongs + bugfixes.
- `SQLDialect.update()` and `SQLDialect.delete()` now take a `Table` object instead of tablename. This change was necessary because `MySQLDialect.delete()` and also other dialects need to access both `sql_shortref` and `sql_fullref` when the table is aliased.
- New method `writing_alias()` added to `SQLDialect`. The method simply returns `table.sql_fullref`. It was added so that dialects like SQLite which don't support `UPDATE/DELETE` on aliased tables can raise exception without reimplementing the whole `update()`/`delete()` methods.
- IMAP adapter was updated to new API (passing `Table` objects instead of tablename). I forgot to do this in the subselect pull request. Beware that this change is untested.
- `Field._rname` and `Field._raw_rname` will now be stored in pickled table metadata files.

### DIFF
--- a/pydal/adapters/firebird.py
+++ b/pydal/adapters/firebird.py
@@ -51,6 +51,7 @@ class FireBird(SQLAdapter):
         return long(self.cursor.fetchone()[0])
 
     def create_sequence_and_triggers(self, query, table, **args):
+        # FIXME: use table._rname instead?
         tablename = table._tablename
         sequence_name = table._sequence_name
         trigger_name = table._trigger_name
--- a/pydal/adapters/firebird.py
+++ b/pydal/adapters/firebird.py
@@ -51,8 +51,7 @@ class FireBird(SQLAdapter):
         return long(self.cursor.fetchone()[0])
 
     def create_sequence_and_triggers(self, query, table, **args):
-        # FIXME: use table._rname instead?
-        tablename = table._tablename
+        tablename = table._rname
         sequence_name = table._sequence_name
         trigger_name = table._trigger_name
         self.execute(query)
--- a/pydal/adapters/ingres.py
+++ b/pydal/adapters/ingres.py
@@ -32,19 +32,18 @@ class Ingres(SQLAdapter):
         # post create table auto inc code (if needed)
         # modify table to btree for performance....
         # Older Ingres releases could use rule/trigger like Oracle above.
-        # FIXME: use table._rname instead in all 3 places?
         if hasattr(table, '_primarykey'):
             modify_tbl_sql = 'modify %s to btree unique on %s' % \
-                (table._tablename,
+                (table._rname,
                  ', '.join(["'%s'" % x for x in table.primarykey]))
             self.execute(modify_tbl_sql)
         else:
-            tmp_seqname = '%s_iisq' % table._tablename
+            tmp_seqname = '%s_iisq' % table._raw_rname
             query = query.replace(self.dialect.INGRES_SEQNAME, tmp_seqname)
             self.execute('create sequence %s' % tmp_seqname)
             self.execute(query)
             self.execute(
-                'modify %s to btree unique on %s' % (table._tablename, 'id'))
+                'modify %s to btree unique on %s' % (table._rname, 'id'))
 
 
 @adapters.register_for('ingresu')
--- a/pydal/adapters/ingres.py
+++ b/pydal/adapters/ingres.py
@@ -32,6 +32,7 @@ class Ingres(SQLAdapter):
         # post create table auto inc code (if needed)
         # modify table to btree for performance....
         # Older Ingres releases could use rule/trigger like Oracle above.
+        # FIXME: use table._rname instead in all 3 places?
         if hasattr(table, '_primarykey'):
             modify_tbl_sql = 'modify %s to btree unique on %s' % \
                 (table._tablename,
--- a/pydal/adapters/oracle.py
+++ b/pydal/adapters/oracle.py
@@ -61,8 +61,7 @@ class Oracle(SQLAdapter):
 
     def create_sequence_and_triggers(self, query, table, **args):
         tablename = table._rname
-        # FIXME: use _rname instead?
-        id_name = table._id.name
+        id_name = table._id._rname
         sequence_name = table._sequence_name
         trigger_name = table._trigger_name
         self.execute(query)
--- a/pydal/adapters/oracle.py
+++ b/pydal/adapters/oracle.py
@@ -61,6 +61,7 @@ class Oracle(SQLAdapter):
 
     def create_sequence_and_triggers(self, query, table, **args):
         tablename = table._rname
+        # FIXME: use _rname instead?
         id_name = table._id.name
         sequence_name = table._sequence_name
         trigger_name = table._trigger_name
@@ -94,8 +95,8 @@ class Oracle(SQLAdapter):
 
     def _build_value_for_insert(self, field, value, r_values):
         if field.type is 'text':
-            r_values[':' + field.sqlsafe_name] = self.expand(value, field.type)
-            return ':' + field.sqlsafe_name
+            r_values[':' + field._rname] = self.expand(value, field.type)
+            return ':' + field._rname
         return self.expand(value, field.type)
 
     def _insert(self, table, fields):
@@ -103,7 +104,7 @@ class Oracle(SQLAdapter):
             r_values = {}
             return self.dialect.insert(
                 table._rname,
-                ','.join(el[0].sqlsafe_name for el in fields),
+                ','.join(el[0]._rname for el in fields),
                 ','.join(
                     self._build_value_for_insert(f, v, r_values)
                     for f, v in fields)
--- a/pydal/adapters/oracle.py
+++ b/pydal/adapters/oracle.py
@@ -60,7 +60,7 @@ class Oracle(SQLAdapter):
         return long(self.cursor.fetchone()[0])
 
     def create_sequence_and_triggers(self, query, table, **args):
-        tablename = table._rname or table._tablename
+        tablename = table._rname
         id_name = table._id.name
         sequence_name = table._sequence_name
         trigger_name = table._trigger_name
@@ -102,13 +102,13 @@ class Oracle(SQLAdapter):
         if fields:
             r_values = {}
             return self.dialect.insert(
-                table.sqlsafe,
+                table._rname,
                 ','.join(el[0].sqlsafe_name for el in fields),
                 ','.join(
                     self._build_value_for_insert(f, v, r_values)
                     for f, v in fields)
                 ), r_values
-        return self.dialect.insert_empty(table.sqlsafe), None
+        return self.dialect.insert_empty(table._rname), None
 
     def insert(self, table, fields):
         query, values = self._insert(table, fields)
--- a/pydal/migrator.py
+++ b/pydal/migrator.py
@@ -51,6 +51,7 @@ class Migrator(object):
 
                 if referenced == '.':
                     referenced = tablename
+                # FIXME: use table._raw_rname instead?
                 constraint_name = self.dialect.constraint_name(
                     tablename, field_name)
                 # if not '.' in referenced \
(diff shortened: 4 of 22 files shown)


gi0baro:
@nextghost some notes:
1) you said:
- `Table._rname` now defaults to quoted `Table._dalname`
- Added `Table._raw_rname` which is the unquoted version of `_rname`

then `_raw_rname` equals `_dalname`? So is it useless?

2) What's the difference between `sqlsafe` and `sqlsafe_alias`?

3) I propose some more developer-firendly names for all these attributes, deprecating the old-ones, but we can discuss of this later.

nextghost:
1. `_raw_rname` is not useless:
```
>>> db.define_table('foo', ...)
>>> db.foo._rname, db.foo._raw_rname, db.foo._dalname
('"foo"', 'foo', 'foo')
>>> db.define_table('bar', ..., rname='test')
>>> db.foo._rname, db.foo._raw_rname, db.foo._dalname
('test', 'test', 'bar')
```
2. Here's an example with aliased table:
```
>>> table.sqlsafe
'alias'
>>> table.sqlsafe_alias
'rname AS alias'
```
3. Suggestions for better names are welcome. The sooner, the better.

gi0baro:
@nextghost ok seems everything legit.
Maybe in this case:

```python
>>> db.define_table('bar', ..., rname='test')
>>> db.foo._rname, db.foo._raw_rname, db.foo._dalname
('test', 'test', 'bar')
```

We can make `._rname` the quoted version of `_raw_rname` (which is the `rname` parameter passed to the constructor?

Another thing: `_tablename` is the same of `_dalname`?

Regarding the names, this is my proposal:
- `_rname` -> `_quoted_name` or `_qname`
- `_raw_rname` -> `_raw_name`
- `sqlsafe()` -> `sql_name()`
- `sqlsafe_alias()` -> `sql_aliased()`

nextghost:
I think that when `rname` constructor argument is provided by the developer, it should be used as is without quoting.

`_tablename` is the active alias. When you call `Table.with_alias()`, the aliased clone will have the same `_dalname` as the original table but a different `_tablename`.

Also note that `_rname`, `sqlsafe` and `sqlsafe_alias` existed for ages and may be used in third-party code. Especially `sqlsafe`.

gi0baro:
Ok summing up:
- let's keep `rname` logic untouched
- i know `sqlsafe` and `sqlsafe_alias` have a long story, in fact I said let's *deprecate* them, not remove them. It's just for the development *sanity*, so we all know by the name of the variables or the methods what are we doing. You can use the `deprecated` decorator or the `warn_of_deprecation` method in the `utils` module.

nextghost:
In that case, I propose these new names:
- `sqlsafe` -> `sql_alias`
- `sqlsafe_alias` -> `sql_from`

gi0baro:
@nextghost why `from`?
`sql_alias` looks good but if the table isn't aliased is not really an *alias*. Maybe `sql_current_name` instead of `sql_name`?

nextghost:
@gi0baro Because `sqlsafe_alias` is almost exclusively used in the `FROM` clause of SQL queries. On MSSQL, even `UPDATE` query has a `FROM` clause specifically for table references in the form of `rname AS alias`.

mdipierro:
@gi0baro @nextghost Are you both in agreement about this PR?

gi0baro:
@mdipierro this is still a *Work In Progress*, @nextghost will update the status when it will be ready for review and mege

gi0baro:
@nextghost I agree with the fact that the name of tables are arguments of a `FROM` clause, but if i see a method named `sql_from` I expect a `FROM` clause returned, not an argument for it :)
Just as a recap, I would like to have:
- a name for `sqlsafe` that reflects the fact the return value is the **current** table name for the sql I'm generating
- a name for `sqlsafe_alias` that reflects the fact the return value is an alias **in an `AS` form** for the sql I'm generating

nextghost:
@gi0baro How about these?
- `sqlsafe` -> `sql_shortref` (It's not exactly a name.)
- `sqlsafe_alias` -> `sql_fullref` (If it was called `sql_aliased`, I'd expect it to return a `True/False` value whether alias is currently in use. SQL server docs usually call this query token a "table reference".)

gi0baro:
@nextghost seems good to me.

nextghost:
That should be all for tables. Now on to refactoring `Field`. Here's my proposal:
- Make `Field.sqlsafe` raise exception when `self._table` is `None`, otherwise keep it as is
- Deprecate `Field.sqlsafe_name` and instead make `Field._rname` default to quoted `Field.name` (and add `_raw_rname`)
- Add a new attribute `longname` which will return `self._table._tablename + '.' + self.name` (or raise exception)
- Refactor places which should be using `sqlsafe` and `longname` but instead build the same string from pieces.

gi0baro:
@nextghost regarding tables, I'm still missing:

- `_rname` -> `_quoted_name` or `_qname`
- `_raw_rname` -> `_raw_name`

When I said to keep the actual `_rname` logic I intended to keep the logic and the Table parameter but to change the internal name everywhere.

Regarding fields, I'm quite ok with everything, except I didn't fully understood what do you mean with "Refactor places which should be using sqlsafe and longname but instead build the same string from pieces".

Generally speaking: I think the "sql" logic should be unified somehow. We still have some things that are handled directly by the objects, some handled by the adapter, some other by the adapters' dialects. The best thing (IMHO) would be to have the "to sql string" logics inside the dialects.

nextghost:
@gi0baro I think that `_rname` (as in "real name") is sufficiently descriptive as it is. Renaming it to `_quoted_name` or `_qname` would be more confusing since in many cases the string won't be quoted. `_raw_rname` should also keep clear semantic connection to `_rname` so that developers won't be confused why is it "raw".

As for refactoring places which should be using `sqlsafe` and `longname`, here's what `SQLAdapater._expand()` currently does:

    def _expand(self, expression, field_type=None, colnames=False, query_env={}):
        if isinstance(expression, Field):
            et = expression.table
            if not colnames:
                table_rname = et.query_alias
                rv = '%s.%s' % (table_rname, expression._rname or (self.dialect.quote(expression.name)))
            else:
                rv = '%s.%s' % (self.dialect.quote(et._tablename), self.dialect.quote(expression.name))
            if field_type == 'string' and expression.type not in ('string', 'text', 'json', 'password'):
                rv = self.dialect.cast(rv, self.types['text'], query_env)
        elif isinstance(expression, (Expression, Query)):
            ...

And here's what it should be doing instead:

    def _expand(self, expression, field_type=None, colnames=False, query_env={}):
        if isinstance(expression, Field):
            if not colnames:
                rv = expression.sqlsafe
            else:
                rv = expression.longname
            if field_type == 'string' and expression.type not in ('string', 'text', 'json', 'password'):
                rv = self.dialect.cast(rv, self.types['text'], query_env)
        elif isinstance(expression, (Expression, Query)):
            ...

gi0baro:
@nextghost ok, I agree with everything.

nextghost:
`Field` name attributes have been refactored as I've proposed above. Some additional changes that I've made in the process:
- Added new method `Field.bind(table)` which takes care of all the boilerplate for setting references to the new parent table.
- `Field.clone()` now unsets references to parent table and database so that `bind()` can do some extra validation.
- TFK foreign keys (multi-field references to legacy tables) were completely broken, that particular branch of migrator code was apparently never tested. They work now, but the code should be merged with normal single-field foreign keys. It's literally the same thing, just with a few extra field names fed to the template. Many dialects don't even have a `'reference TFK'` template string. I've also added a unit test for TFK foreign keys which skips any adapter that does not have the template string. Remove the skip condition after you merge TFK with normal foreign keys.
- I took the liberty to fix issue #430

gi0baro:
@nextghost do you have any updates on this? any ETA?

nextghost:
@gi0baro I need your input on the remaining FIXMEs in the changeset. When they're resolved, it should be ready for merge.

pydal/adapters/firebird.py
```diff
@@ -51,6 +51,7 @@ def lastrowid(self, table):
         return long(self.cursor.fetchone()[0])
 
     def create_sequence_and_triggers(self, query, table, **args):
+        # FIXME: use table._rname instead?
```
gi0baro: Yes (or alternatively, `sql_shortref`?)

pydal/adapters/ingres.py
```diff
@@ -32,6 +32,7 @@ def create_sequence_and_triggers(self, query, table, **args):
         # post create table auto inc code (if needed)
         # modify table to btree for performance....
         # Older Ingres releases could use rule/trigger like Oracle above.
+        # FIXME: use table._rname instead in all 3 places?
```
gi0baro: Yes (or alternatively, `sql_shortref`?)

pydal/adapters/oracle.py
```diff
@@ -60,7 +60,8 @@ def lastrowid(self, table):
         return long(self.cursor.fetchone()[0])
 
     def create_sequence_and_triggers(self, query, table, **args):
-        tablename = table._rname or table._tablename
+        tablename = table._rname
+        # FIXME: use _rname instead?
```
gi0baro: Yes (or alternatively, `sql_shortref`?)

pydal/migrator.py
```diff
@@ -51,6 +51,7 @@ def create_table(self, table, migrate=True, fake_migrate=False,
 
                 if referenced == '.':
                     referenced = tablename
+                # FIXME: use table._raw_rname instead?
```
gi0baro: Yes.

pydal/migrator.py
```diff
@@ -111,18 +112,19 @@ def create_table(self, table, migrate=True, fake_migrate=False,
...
                     if referenced == tablename:
-                        real_referenced = db[referenced].sqlsafe
+                        real_referenced = db[referenced]._rname
                     else:
                         real_referenced = (
-                            referenced in db and db[referenced].sqlsafe or
+                            referenced in db and db[referenced]._rname or
                             referenced)
                     rfield = db[referenced]._id
+                    # FIXME: use field._raw_rname for index_name?
```
gi0baro: Yes.

pydal/migrator.py
```diff
@@ -157,6 +159,7 @@ def create_table(self, table, migrate=True, fake_migrate=False,
                     else:
                         schema = parms[0]
                     ftype = "SELECT AddGeometryColumn ('%%(schema)s', '%%(tablename)s', '%%(fieldname)s', %%(srid)s, '%s', %%(dimension)s);" % types[geotype]
+                    # FIXME: use table._raw_rname instead?
```
gi0baro: Yes.

pydal/migrator.py
```diff
@@ -353,17 +362,16 @@ def migrate_table(self, table, sql_fields, sql_fields_old, sql_fields_aux,
                    ftype.startswith('geometry'):
                     geotype, parms = ftype[:-1].split('(')
                     schema = parms.split(',')[0]
+                    # FIXME: use table._raw_rname instead?
```
gi0baro: Yes.

pydal/adapters/oracle.py
nextghost: `Field` objects have no `sql_shortref` attribute.

nextghost:
This pull request is now ready to be merged. It also fixes issues #57, #343 and #430. I'll add a summary of the most important changes into the opening comment tomorrow.

gi0baro:
@nextghost great! I'll wait for the opening update, then will merge this.

nextghost:
@gi0baro You should also review the last commit (9d5c38a) because I've noticed at the last moment that field migrations had no support for `Field._rname` whatsoever. I even had to add extra field items to the pickled metadata files so that you could actually drop fields with distinct rname.

gi0baro:
@nextghost everything looks good to me.
Unless @mdipierro or @niphlod have any concerns, I will merge this by the end of the week.

Issue #442, opened 2017-01-19
Error inserting to table with UUID for Primary Key


peregrinius:
```
Traceback (most recent call last):
  File "/opt/codebase/pydal/pydal/adapters/postgres.py", line 109, in lastrowid
    return long(self.cursor.fetchone()[0])
ValueError: invalid literal for int() with base 10: '5024590b-322b-45b2-9343-ad989e334cea'
```

PR #443, opened 2017-01-19
[WIP] UUID as primary key

james-burke-codes:
Resolution to issue #442 

This pull request will enable Postgresql users to use a UUID as a primary key. Currently when trying to do an insert the code expects an integer to be returned resulting in a ValueError exception.

### DIFF
--- a/pydal/adapters/postgres.py
+++ b/pydal/adapters/postgres.py
@@ -105,7 +105,11 @@ class Postgre(with_metaclass(PostgreMeta, SQLAdapter)):
 
     def lastrowid(self, table):
         if self._last_insert:
-            return long(self.cursor.fetchone()[0])
+            pk = self.cursor.fetchone()[0]
+            try:
+                return long(pk)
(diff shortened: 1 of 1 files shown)


mdipierro:
@gi0baro I am fine with this. If you are too, I would merge it.

gi0baro:
@mdipierro I'm not really happy with the implementation. IMHO if we really want to support this we should plan a complete implementation.

gi0baro:
@mdipierro also, this should still be considered a *Work in progress*, see the title.

mdipierro:
I agree. In the case of Mongodb, my original approach was to map UUIDs in long integers and vice versa.

chthor:
+1 for string id support. [My application](https://github.com/web2py/pydal/issues/607) requires a string id field and it looks like currently there's no way I can make that work.

mdipierro:
I am going to decline this. It is a good idea but requires a lot more changes than this one.

PR #446, opened 2017-02-03
Bugfix in firebird adapter

josejachuf:
Bugfix to the problem introduced in version 16.06


gi0baro:
Implemented with d455335

PR #448, opened 2017-03-01
Web2py compatibility bugfixes

nextghost:
My recent contributions have apparently broken a lot of things in Web2py so I should fix that. This pull request fixes:
- `SQLFORM.factory()` errors.
- Migrations getting triggered on every request due to minor metadata differences which never actually migrate anything at all.
- Caching `Rows` objects to disk.

`SQLTABLE` class is also broken by table/field name refactoring but that needs to be fixed in Web2py. I'll open another pull request for that.

### DIFF
--- a/pydal/migrator.py
+++ b/pydal/migrator.py
@@ -429,7 +429,7 @@ class Migrator(object):
                     drop_expr % (table._rname, key_tmp)
                 ]
                 metadata_change = True
-            elif sql_fields[key]['type'] != sql_fields_old[key]['type']:
+            elif sql_fields[key] != sql_fields_old[key]:
                 sql_fields_current[key] = sql_fields[key]
                 metadata_change = True
 
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -745,10 +745,12 @@ class SQLAdapter(BaseAdapter):
         key = self.uri + '/' + sql
         key = hashlib_md5(key).hexdigest()
         args = (sql, fields, attributes, colnames)
-        return cache_model(
+        ret = cache_model(
             key,
             lambda self=self, args=args: self._select_aux(*args),
             time_expire)
+        ret._restore_fields(fields)
+        return ret
 
     def select(self, query, fields, attributes):
         colnames, sql = self._select_wcols(query, fields, **attributes)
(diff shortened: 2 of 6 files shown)


nextghost:
Hmm, any suggestions what to do with `Row` objects failing to pickle on MSSQL? Is there some way to fix that or should I just disable the test for MSSQL adapters?

nextghost:
@gi0baro Actually, that's my pickling test that's failing now. But it seems to be caused by class nesting in the ODBC driver so I'll just disable part of the test.

That `COALESCE()` test failure is weird because the last good test run was on commit that was basically identical to the first that started failing. Did anybody contact Appveyor about this?

nextghost:
@gi0baro OK, this is ready to merge whenever you like.

pydal/migrator.py
```diff
@@ -429,7 +429,7 @@ def migrate_table(self, table, sql_fields, sql_fields_old, sql_fields_aux,
                     drop_expr % (table._rname, key_tmp)
                 ]
                 metadata_change = True
-            elif sql_fields[key]['type'] != sql_fields_old[key]['type']:
```
gi0baro: @nextghost are you sure about this?

nextghost: Yes. If the metadata are still not identical at this point, the field can either be safely fake-migrated or it needs to be fixed manually anyway. Or do you prefer triggering migrations which do nothing on every single Web2py request that uses the table?

gi0baro: @nextghost ok, but if the issue is that you're migrating on every request because this `if` condition is triggered, then the problem is not the condition itself but the fact that the metadata are different even if they should be identical. I mean: if the problem is that `sql_fields[key]['type'] != sql_fields_old[key]['type']` is `True`, shouldn't we fix the fact that the two types should match instead of dropping the condition? Or am I missing something?

nextghost: No, we're migrating on every request because the actual difference in metadata does **NOT** trigger this condition. But it does trigger a call to `migrate_table()` in the parent migrator method. And I'm not dropping the condition. Look at the next line in the diff.

gi0baro: Ok, sounds right. Gonna merge this later today.

PR #449, opened 2017-03-01
Init adapters properly even when reusing idle connection from pool

nextghost:
Some adapters are not properly initialized when they reuse idle connection from pool. In case of PostgreSQL, this actually results in unnecessary migrations of `json` fields back and forth between `TEXT` and `JSON` SQL types because the adapter will not check whether it can use dialects with native JSON support.

This pull request introduces new `after_reconnect()` adapter method which will be executed every time the connection is assigned to a new adapter object, no matter whether the connection was freshly created or reused from connection pool.

### DIFF
--- a/pydal/adapters/informix.py
+++ b/pydal/adapters/informix.py
@@ -31,7 +31,7 @@ class Informix(SQLAdapter):
     def connector(self):
         return self.driver.connect(self.dsn, **self.driver_args)
 
-    def after_reconnect(self):
+    def after_first_reconnect(self):
         self.dbms_version = int(self.connection.dbms_version.split('.')[0])
 
     @with_connection_or_raise
--- a/pydal/adapters/informix.py
+++ b/pydal/adapters/informix.py
@@ -31,7 +31,7 @@ class Informix(SQLAdapter):
     def connector(self):
         return self.driver.connect(self.dsn, **self.driver_args)
 
-    def after_connection(self):
+    def after_reconnect(self):
         self.dbms_version = int(self.connection.dbms_version.split('.')[0])
 
     @with_connection_or_raise
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -70,7 +70,7 @@ class Mongo(NoSQLAdapter):
         conn.commit = lambda: None
         return conn
 
-    def after_reconnect(self):
+    def after_first_reconnect(self):
         #: server version
         self._server_version = self.connection.command(
             "serverStatus")['version']
--- a/pydal/adapters/mongo.py
+++ b/pydal/adapters/mongo.py
@@ -70,7 +70,7 @@ class Mongo(NoSQLAdapter):
         conn.commit = lambda: None
         return conn
 
-    def after_connection(self):
+    def after_reconnect(self):
         #: server version
         self._server_version = self.connection.command(
             "serverStatus")['version']
--- a/pydal/adapters/postgres.py
+++ b/pydal/adapters/postgres.py
@@ -101,7 +101,6 @@ class Postgre(with_metaclass(PostgreMeta, SQLAdapter)):
     def after_connection(self):
         self.execute("SET CLIENT_ENCODING TO 'UTF8'")
         self.execute("SET standard_conforming_strings=on;")
-        self._config_json()
 
     def lastrowid(self, table):
         if self._last_insert:
@@ -141,7 +140,7 @@ class Postgre(with_metaclass(PostgreMeta, SQLAdapter)):
 class PostgrePsyco(Postgre):
     drivers = ('psycopg2',)
 
-    def _config_json(self):
+    def after_reconnect(self):
         use_json = self.driver.__version__ >= "2.0.12" and \
             self.connection.server_version >= 90200
         if use_json:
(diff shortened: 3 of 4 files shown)


gi0baro:
@nextghost I'm missing the point here.
Adapter init has nothing to do with connection pooling and re-using.

I don't get why you want to execute commands you already executed just because you're re-using an existing connection.

nextghost:
@gi0baro Which part of "This bug screws up dialect selection on PostgreSQL" did you not understand?

    args = dict(uri='postgres://user:pass@localhost/pydal', pool_size=10)
    for cycle in ['First', 'Second']:
        db = DAL(**args)
        print('{0}: {1}'.format(cycle, type(db._adapter.dialect).__name__))
        db.close()

On current master:

    First: PostgreDialectJSON
    Second: PostgreDialect

After applying this patch:

    First: PostgreDialectJSON
    Second: PostgreDialectJSON

gi0baro:
@nextghost I needed an example. Will inspect this better, 'cause setting the encoding on every re-connection is not a *clean solution* IMHO.

nextghost:
Read the patch FFS. `after_reconnect()` does **NOT** set encoding.

nextghost:
@gi0baro Read the patch FFS. `after_reconnect()` does **NOT** set encoding.

gi0baro:
@nextghost check #451, should solve the issue without slowing down every reconnection. If you agree, will merge #451 and close this.

nextghost:
@gi0baro That's an awful lot of complexity just to avoid calling this amount of code:

    def after_reconnect(self):
        pass

The only adapters that overload this method will be:
- Postgres
- Mongo
- Informix

The method doesn't exist in current master.

gi0baro:
@nextghost I agree about the code complexity compared to yours, but there's an important difference:
- your code re-initialize the dialect and the parser/representer **every time** you get a connection from the pool
- mine runs the code just once per adapter instance

My point is not to decrease the performance on postgres just because of this issue.

nextghost:
@gi0baro Unless `after_reconnect()` raises exception, my code will also run only once per adapter instance. If it does raise exception, then you need to redo the initialization anyway. The only difference is that your version will backtrack all the way back to `DAL.__init__()`, waste a connection attempt and `sleep()` for 1 second while mine will just grab another connection from the pool right away.

gi0baro:
@nextghost ehm.. no.

```python
self.connection = POOLS[uri].pop()
GLOBAL_LOCKER.release()
try:
    if self.check_active_connection:
        self.test_connection()
    print('foo')
    break
except:
    pass
```

*foo* will be printed every time you get an alive connection out of the pool. Test it with pre-fork or threads.

nextghost:
@gi0baro How exactly would the same adapter instance take more than one usable connection out of the pool (unless you call `reconnect()` on it manually)? That loop is there only to skip connections which are already dead.

gi0baro:
@nextghost 'cause pyDAL is not web2py, and you can use it in different ways rather than instantiate the whole DAL on every request. pyDAL is used by other frameworks that use pre-forking and the ability to use it in a multi-threaded environment is a feature. Check this scheme:

```
- main process
  - static POOLS
  - DAL instance/adapter instance
    - thread1
      - connection 1
        - cursor 1
        - cursor 2
      - connection 2
        - cursor 1
    - thread 2
      - connection 1
        - cursor 1
      - connection 2
        - cursor 1
        - cursor 2
```

if every thread performs it's own `connect`, your `after_reconnect` is performed 4 times. That should be correct if the contents of `after_reconnect` should be performed on every reconnection, but the configuration of the adapter (json stuff on postgres) should be performed just one time on the adapter, independently on how many times you re-use a connection or if a connection fails. You got it now?

nextghost:
Ah, OK, I didn't notice that `adapter.connection` is a thread-local property. But in that case, `after_reconnect()` should still be called from `ConnectionPool.reconnect()` under instance-level lock and then disabled. If the adapter instance gets created with `do_connect=False` and then multiple threads try to `reconnect()` at the same time, both our versions might lead to race conditions in the future (if somebody adds code to the callback that doesn't produce the same result in all threads).

Also, `BaseAdapter.index_expander()` breaks thread safety.

nextghost:
@gi0baro Ah, OK, I didn't notice that `adapter.connection` is a thread-local property. But in that case, `after_reconnect()` should still be called from `ConnectionPool.reconnect()` under instance-level lock and then disabled. If the adapter instance gets created with `do_connect=False` and then multiple threads try to `reconnect()` at the same time, both our versions might lead to race conditions in the future (if somebody adds code to the callback that doesn't produce the same result in all threads).

Also, `BaseAdapter.index_expander()` breaks thread safety.

nextghost:
New version of my solution:
- thread-safe
- executed once per adapter instance even in multi-threaded environment

gi0baro:
@nextghost I've updated my own too, added locking and created a mixin class for mocking the reconnection.
Even if my solution has more code, it avoids the lock and the if condition on every reconnection.

gi0baro:
And `SQLAdapter.index_expander()` is intended to be used only in migrations code, so it's ok to not be thread safe.

nextghost:
@gi0baro If you insist on method swapping for performance optimization, here's a version that isn't ridiculously overengineered. I also did some performance tests and this version has no measurable performance impact.

The previous version had ~20 microseconds of overhead on my Core i5 laptop when connection check was turned off (tested on 100,000 reconnects, either with 10 threads doing 10,000 reconnects each, or with 100 threads doing 1,000 reconnects each). That's about 20% slowdown on Postgre, under 8% on MySQL. With connection check turned on, there was no measurable difference.

gi0baro:
@nextghost sorry man, but I still prefer that adapters implement the mocking rather than touching the connection. I'm gonna merge the other one.

Also, let me know if you have any additions on-going, otherwise I'm gonna build a new release during the week.

PR #453, opened 2017-03-05
added tests for py36

ilvalle:


gi0baro:
@ilvalle can we also drop python 3.3 from the matrix?

ilvalle:
I'd keep py33 in the loop for a little more

gi0baro:
@ilvalle ok

PR #457, opened 2017-04-05
added missing "oracle" in adapters.register_for()

leonelcamara:
Fixes web2py/web2py#1599


jasonphillips:
This is one of a couple of completely breaking bugs that were strangely allowed to creep unnoticed into the oracle adapter in commits over past year. Would love to see this merged so I can submit a fix for the other one. And I'm interested in anyone's knowledge of setting up a complete test for this adapter, so we won't have to worry about breaking changes in stable releases.

PR #464, opened 2017-05-03
Add full Oracle testing to Travis, fix the adapter

jasonphillips:
Since so many errors and fundamental issues seem to have crept into the Oracle adapter over time, I took on the task of setting up the automated Travis tests to include Oracle. 

> **Note**:  using Oracle on Travis requires a download of the free Express Edition Oracle platform during container build, since it cannot be redistributed, so to make it work (via [this handy project](https://github.com/cbandy/travis-oracle)) you need to add two environment variables to Travis directly: a username / password for a (free) oracle.com account that has accepted their license agreement for this product. I can email that username / password, and once added to the main `pydal` Travis account, the build will succeed.  [See the successful build here.](https://travis-ci.org/jasonphillips/pydal/builds/228438601)

Once I started running the tests on Oracle, a considerable number of fundamentally breaking issues cropped up everywhere--including everything from broken python syntax to badly handled sql generation and types--but after many adjustments and improvements to the adapter, dialect, parser, etc, the entire test suite now runs successfully (except for a handful of tests that are not appropriate for Oracle, like the Time type it doesn't support, which I toggled off for this adapter in `tests/sql.py`). 

The biggest challenge for Oracle was properly handling case-sensitive table names and field names (when not using entity_quoting=False), which requires that they be quoted correctly _everywhere_. I did need to add a few additional switches to the main `migrator.py` file to make those additional quotes possible in places where it didn't currently invoke `dialect.quote` on an `rname`. Those I wrapped in an if-oracle condition for now, but perhaps it would be best in the long run to automatically invoke the quote function on rnames for these additional cases for all adapters. 

Everything now works smoothly--even complex things like correct results paging for all the test instances, with far greater stability than the prior Oracle query-paging syntax--and I'm hoping that moving it into the automated tests can finally make Oracle a first-tier citizen again for pydal, given that it had drifted into severe problems over time.


jasonphillips:
Once again, view the full successful build / test results here on Travis:

https://travis-ci.org/jasonphillips/pydal/builds/228438601

The build under this account will fail until the private environment variables mentioned above are added manually.

tests/sql.py
```diff
@@ -224,7 +224,10 @@ def testRun(self):
         for ft in ['string', 'text', 'password', 'upload', 'blob']:
             db.define_table('tt', Field('aa', ft, default=''))
             self.assertEqual(db.tt.insert(aa='x'), 1)
-            self.assertEqual(db().select(db.tt.aa)[0].aa, 'x')
+            if not (IS_ORACLE and (ft=='text' or ft=='blob')): 
+                # only verify insert for LOB types in oracle;
+                # select may throw errors in test
```
jasonphillips: Not checking the CLOB/LOB return from Oracle is necessary only in the Travis container environment -- a memory allocation error from Oracle's end is thrown that does not occur if run against a production database.

tests/sql.py
```diff
@@ -553,7 +558,7 @@ def testGroupByAndDistinct(self):
         self.assertEqual(len(result), 1)
         self.assertEqual(tuple(result.response[0]), ('3', 3))
         result = db().select(db.tt.aa, db.tt.bb.sum(),
-                             groupby=db.tt.aa, limitby=(0,3))
+                             groupby=db.tt.aa, orderby=db.tt.aa, limitby=(0,3))
```
jasonphillips: This test was inconsistent due to lacking an `orderby` declaration.

mdipierro:
Any idea why the appveyor/pr test fails?

jasonphillips:
It looks from the build history on appveyor as if the build has been failing for at least a few commits before this PR, is that correct? I checked out the test line that is currently failing there, and it's one of the ones I had to skip for oracle, so it might need to be skipped for mssql as well:

```python
        db.tt.insert(aa='xx')
        db.tt.insert(aa='xx', bb='yy')
        db.tt.insert(aa='xx', bb='yy', cc='zz')
        if not IS_ORACLE:
            # empty string is treated as null
            db.tt.insert(aa='xx', bb='yy', cc='zz', dd='')
        result = db(db.tt).select(db.tt.dd.coalesce(db.tt.cc, db.tt.bb, db.tt.aa))
        self.assertEqual(result.response[0][0], 'xx')
        self.assertEqual(result.response[1][0], 'yy')
        self.assertEqual(result.response[2][0], 'zz')
        if not IS_ORACLE:
            self.assertEqual(result.response[3][0], '')  # <-- this line fails on mssql as well
        db.tt.drop()
```

The test expects coalesce to return the value of the `dd` column since it's an empty string `''` rather than a `null`, but Oracle doesn't have a distinction between empty string and null, so coalesce will skip that value and give the next ('zz') if run. That's why I comment for Oracle. 

It looks from the appveyor message as if mssql handles the coalesce as expected (returning the value of the dd column) but the adaptor actually returns that as type `None` rather than `''`, so the test fails.

Appveyor:

```
  File "tests\sql.py", line 613, in testCoalesce
    self.assertEqual(result.response[3][0], '')
AssertionError: None != ''
----------------------------------------------------------------------
```

I haven't looked at the mssql adapter in detail yet, but would guess that we could either (1) skip that test for it, or (2) check into adjusting its adapter to return an empty string rather than a null when a text column has an empty string.

jasonphillips:
Looking a bit deeper into the appveyor history (which again is a little off-topic from this commit), I can see that the build started failing back in January, between these two builds:  

 - succeeded on Jan 23 https://ci.appveyor.com/project/web2py/pydal/build/1.0.476
 - failed on Jan 30 https://ci.appveyor.com/project/web2py/pydal/build/1.0.477/job/kulaw78bx0q4dgjd

Those are essentially testing identical code, so there has to be a change in a required library between those two dates -- and I think pypyodbc is almost certainly the culprit, since it saw a number of significant changes and commits in exactly that timeframe:

https://github.com/jiangwen365/pypyodbc/commits/master

mdipierro:
I would like to merge this but I am concerned about the pyodbc error you refer to. Is your understand that that those changes that broke CI are a bug and should be fixed by them, not us?

jasonphillips:
Correct, and at the very least, the Appveyor build failures began some time ago, and are currently happening on the latest build apart from this PR, so the code here doesn't alter anything on that point. I tried to figure out what the change back in January might have been that is causing only one version of mssql to fail during tests (because the others are all passing), but the only clue I could find was indeed the pyodbc changes that preceded the day that test started failing.

Anyhow, in case you do get to the point of merging this, I just sent you the details of those 2 environment variables that would need to be set in Travis.

mdipierro:
You have done lots of work here and I really want to merge it but, it conflicts with previous patches and it seems to fail all the Oracle tests. Can you please reconcile, make sure that all pre-existing tests pass, and disable new tests that you added if they do not pass travis?

jasonphillips:
After merging and solving conflicts, the tests are now dying early due to a peculiar change they just published in cx_Oracle v6, which was released recently (this error: https://github.com/oracle/python-cx_Oracle/issues/67 -- and their explanation isn't sufficient IMO). Evidently calling `.close()` on a cx_Oracle connection is no longer functional, so I'm going to lock the library version in the Travis tests to their latest 5.x stable cx_Oracle to see if that passes.

jasonphillips:
Okay this has resolved conflicts and now passes all travis tests (full results: https://travis-ci.org/jasonphillips/pydal/builds/268671504 - recall that they'll fail here on the main pydal until the travis environment variables have been set, as mentioned in first comment above).

mepdr:
@mdipierro @jasonphillips what is the status of this PR ? It would be very useful to see those changes applied into master

mdipierro:
Will check today. Lots of conflicts.

mepdr:
@mdipierro have you tried to solve the conflicts?

jasonphillips:
I merged and fixed basic conflicts, and now it's failing on the relatively new test for blob bytes added in https://github.com/web2py/pydal/pull/520

I'll see if I can find a chance to debug from here soon.

jasonphillips:
After a few adjustments around character encoding, the tests are all passing again. See the results here:

https://travis-ci.org/jasonphillips/pydal/builds/366578216

(Remember that the tests on this repo, rather than on my fork, will fail on Travis until the Oracle account environment variables are added to your Travis build, as detailed above in the discussion. These are for a free Oracle account that has accepted their license agreement to download the Oracle lite version that supports these tests. I can email those - I'm jphillips(at)ua.edu when you need to obtain them).

mdipierro:
I finally merged this and I am sorry it took so long. It still does not pass travis. https://travis-ci.org/github/web2py/pydal/jobs/688170825 I am hoping you can with that. :-)

PR #465, opened 2017-05-16
Fix get in pydal.helpers.classes Reference class calling __getattr__ …

emman27:
…with 3 variables instead of 2 as per signature

### DIFF
--- a/pydal/helpers/classes.py
+++ b/pydal/helpers/classes.py
@@ -213,19 +213,19 @@ class Reference(long):
                 "reference: %s %d" % (self._table, long(self))
             )
 
-    def __getattr__(self, key):
+    def __getattr__(self, key, default=None):
         if key == 'id':
             return long(self)
         if key in self._table:
             self.__allocate()
         if self._record:
             # to deal with case self.update_record()
-            return self._record.get(key, None)
+            return self._record.get(key, default)
         else:
             return None
 
     def get(self, key, default=None):
-        return self.__getattr__(key) or default
+        return self.__getattr__(key, default)
 
     def __setattr__(self, key, value):
         if key.startswith('_'):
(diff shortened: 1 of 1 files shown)


mdipierro:
Can you explain the rationale for this change?

emman27:
@mdipierro if you look at the method signature:
```def __getattr__(self, key):```.
However, the current code calling this method is
```return self.__getattr__(key, default)```
Since `self` is automatically appended onto the method call, this raises a TypeError.
The code for `__getattr__` is in fact fine and does not handle the default case,, whereas for `get` which was changed in this PR, the default case is incorrectly passed on instead of directly handled

**Comment by mdipierro** [comment]
I think a better solution than is to add a default option to 
```
base.py:    def __getattr__(self, key): to __getattr__(self, key, default=None)
objects.py:    def __getattr__(self, k): to __getattr__(self, k, default=None)
```
because help(getattr) allows of a default option. Also we do not want to get the default is the attribute is defined and its value is 0 or ''. 

If I agree, may I ask you to make this change? else I can do it.

**Comment by emman27** [comment]
You're right on that count. However I think the fix that needs to be made is in line 216 (onwards) of https://github.com/web2py/pydal/blob/master/pydal/helpers/classes.py, where it should read, letting the underlying default be handled by the `get` method in `objects.py` which already works fine in handling defaults in the case of KeyErrors (I am assuming here that `self._record` refers to a `Row` object, correct me if I am wrong)
```
def __getattr__(self, key, default=None): # Change made here to include default in method signature
        if key == 'id':
            return long(self)
        if key in self._table:
            self.__allocate()
        if self._record:
            # to deal with case self.update_record()
            return self._record.get(key, default) # Change made here to switch None to default
        else:
            return None
```

mdipierro:
Yes. Please make the change. I will also change all the other __getattr__ for uniformity.

PR #469, opened 2017-06-16
Make DAL support jsonb for postgres

leonelcamara:


leonelcamara:
This will let postgres users use jsonb instead of json if they prefer it.

PR #475, opened 2017-06-29
Update mongo.py

ftcmnc:
Correction for invalid list:reference representer
Now accepts array of foreign keys

### DIFF
--- a/pydal/representers/mongo.py
+++ b/pydal/representers/mongo.py
@@ -47,5 +47,7 @@ class MongoRepresenter(NoSQLRepresenter):
 
     @for_type('list:reference')
     def _list_reference(self, value):
+        if not isinstance(value, list):
+            return self.adapter.object_id(value)
         values = self._represent_list(value)
         return list(map(self.adapter.object_id, values))


gi0baro:
@ftcmnc can't see the point of this change. `Representer._represent_list` already parse the lists.

ftcmnc:
@gi0baro I believe this was an issue while using web2py 2.14.6 , it seams to work correctly using pydal 17.07 in web2py 2.15.2

ftcmnc:
Hi @gi0baro after all the problem still persists in pydal 17.07 using web2py 2.15.2
This happens when you have a list of ObjectIds.. if you try to update that list using a form, it passes something like this: `"industries": [ObjectId("578b619a2ba3d3729a442c82")]` to something like this `"industries": [ [ObjectId("578b619a2ba3d3729a442c82")] ]`

mdipierro:
Can you please comment about my comment?

gi0baro:
@ftcmnc @mdipierro seems a web2py problem, not pydal related.

mdipierro:
Maybe. I still do not know what does this happen? We need some come to reproduce the issue. For now the proposed solution does not look the right one.

PR #492, opened 2017-09-11
Fix missing urllib.unquote for Python 3.x

wasare:
Minor fix to compatibility with Python 3.x .

### DIFF
--- a/pydal/base.py
+++ b/pydal/base.py
@@ -129,11 +129,9 @@ import socket
 import threading
 import time
 import traceback
-import urllib
-from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


mdipierro:
Thank you for the fix. Can I ask for a change? Can you moved you import for unquote in _compat.py and import unquote from _compat?

pydal/base.py
```diff
@@ -129,11 +129,9 @@
 import threading
 import time
 import traceback
-import urllib
-from uuid import uuid4
```
wasare: This line was delete by mistake, but fixed in the next commit.

mdipierro:
Thank you!

PR #499, opened 2017-11-09
use all caps for COUNT for consistency with the other operations

leonelcamara:


leonelcamara:
Build keeps failing because travis doesn't seem to have postgis... I don't see any of the CI failings to be related with this pull request.

PR #517, opened 2018-02-07
DAL import_table_definitions PY3 fix

maxcrystal:


pydal/base.py
```diff
@@ -530,9 +530,12 @@ def import_table_definitions(self, path, migrate=False,
...
                             notnull=value.get('notnull', False),
                             unique=value.get('unique', False)))
-                        for key, value in sql_fields.iteritems()
+                        for key, value in iteritems(sql_fields)
                     ]
-                    mf.sort(lambda a, b: cmp(a[0], b[0]))
+                    if PY2:
+                        mf.sort(lambda a, b: cmp(a[0], b[0]))
+                    else:
+                        mf.sort(key=lambda a: a[0])
```
mdipierro: can we just make it mf.sort(key=lambda a: a[0]) for both py2 and py3?

maxcrystal: Key sorting works since py2.4, so it's possible I think.

PR #531, opened 2018-04-11
Reconstructing query from dict not working

### LINKED ISSUE #530: Reconstructing query from dict not working
Is this not working any more ?

```python
>>> q = (db.person.id == 1)
>>> db(q.as_dict(flat=True)).select()
  File "<string>", line unknown
SyntaxError: Operator not supported: eq
```

That is in a web2py shell, I mean is the same thing passing around the query in the session as a dict, is not ?

ybenitezf:
- Updating the code in `pydal/objects` for reconstructing a `Query` object from a dict.
- Fix for `uploadfs.open` with expect a `unicode` not a `str`

close #530

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -46,12 +46,12 @@ DEFAULT_REGEX = {
     'id':      '[1-9]\d*',
     'decimal': '\d{1,10}\.\d{2}',
     'integer': '[+-]?\d*',
-    'float':   '[+-]?\d*(\.\d*)?', 
+    'float':   '[+-]?\d*(\.\d*)?',
     'double':  '[+-]?\d*(\.\d*)?',
     'date':    '\d{4}\-\d{2}\-\d{2}',
     'time':    '\d{2}\:\d{2}(\:\d{2}(\.\d*)?)?',
     'datetime':'\d{4}\-\d{2}\-\d{2} \d{2}\:\d{2}(\:\d{2}(\.\d*)?)?',
-    }             
+    }
 
 class Row(BasicStorage):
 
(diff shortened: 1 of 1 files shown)


ybenitezf:
¿any one?

pydal/objects.py
```diff
@@ -2172,14 +2172,15 @@ def build(self, d):
                              d.get("second", None))
         left = right = built = None
 
-        if op in ("AND", "OR"):
+        op = op.upper()
```
mdipierro: I understand this fix. But why change AND -> _AND?

ybenitezf: When the query is pickle to the session the operators *and*, *or* and *not* are prefixed with `_`, so: `_AND`, `_OR` and `_NOT`

```python
In [1]: q = db(db.auth_user.id > 0) & (db.auth_user.id < 10)

In [2]: q
Out[2]: <Query (("auth_user"."id" < 10) AND <Set ("auth_user"."id" > 0)>)>

In [3]: q.as_dict()
Out[3]: 
{'_db': <DAL uri="sqlite://storage.sqlite">,
 'db': <DAL uri="sqlite://storage.sqlite">,
 'first': <Query ("auth_user"."id" < 10)>,
 'ignore_common_filters': False,
 'op': <bound method SQLiteDialect._and of <pydal.dialects.sqlite.SQLiteDialect object at 0x7f897554ff90>>,
 'optional_args': {},
 'second': <Set ("auth_user"."id" > 0)>}

In [4]: q.as_dict(flat=True)
Out[4]: 
{'first': {'first': {'fieldname': 'id', 'tablename': 'auth_user'},
  'ignore_common_filters': False,
  'op': 'lt',
  'optional_args': {},
  'second': 10},
 'ignore_common_filters': False,
 'op': '_and',
 'optional_args': {}}
```

Note to that the `<` is not prefixed with `_`. Actually all the operatos are lowercase, i just make then upper there for the sake of not making to many changes.

ybenitezf:
There is a problem in this, LTE (`<=`) and GTE (`>=`) are not processed correctly

pydal/objects.py
mdipierro: Sorry I am slow I still do not understand where do the '_' comes from?

mdipierro:
This is a very old ticket and I was not able to understand or reproduce the problem. I am going to close it but I may very well be wrong. The example uses a syntax
`db(db.auth_user.id > 0) & (db.auth_user.id < 10)`
that was never supported. The correct syntax is
`db((db.auth_user.id > 0) & (db.auth_user.id < 10))`
or
`db(db.auth_user.id > 0)(db.auth_user.id < 10)`

Issue #461, opened 2017-04-20
skip parameter should be int and not long type


jurkovicb:
When using REST interface in web2py through parse_as_rest offset and skip parameters should be int and not long.

limitby in rest.py are assigned as long numbers.

in mongo.py (line 297) also limitby[0] should be cast as integer.

`limitby_skip, limitby_limit = int(limitby[0]), int(limitby[1]) - 1`

Otherwise pymongo/cursor.py report error (line 125).

PR #532, opened 2018-04-11
db.parse_as_rest on python 3

JusticeN:
python3 doesn't know the long keyword. it raise a NameError when i run db.parse_as_rest(patterns, args, vars) on python 3. Here ist proposition to solve it.
see issue #461


BuhtigithuB:
Please see #533, I think it the way to go. Please close the web2py issue you open about the same topic.

PR #553, opened 2019-02-01
fix chronic migration issue #554

richboss:
I am new to the code base. Confident that this fixes the reported issue, but quite unsure about possible side effects.


mdipierro:
nice catch!

PR #556, opened 2019-02-07
Set version number to 19.02

kvanzuijlen:
Changed version number to reduce missteps. It should be clear to developers what the current version is. Currently it's nowhere to be found, not in setup.py (which uses this version number), not in CHANGES (which is outdated) and not in README.md (which states that version 17.11 is the latest version on PyPi).

It would be great if this was released by the end of this month (that's why I changed the version number to 19.02).

### DIFF
--- a/pydal/__init__.py
+++ b/pydal/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '17.11'
+__version__ = '19.02'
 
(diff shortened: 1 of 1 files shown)


kvanzuijlen:
And please release it to PyPi

PR #561, opened 2019-03-05
experimental json expressions and queries support

leonelcamara:
I'm not sure about the name I've given to stuff so feel free to comment on that or on anything else. I've been using json heavily and being able to query json using the DAL is something I've wanted for a long time.

### DIFF
--- a/pydal/dialects/postgre.py
+++ b/pydal/dialects/postgre.py
@@ -249,10 +249,7 @@ class PostgreDialectJSON(PostgreDialect):
             second['precision'], second['options'])
 
     def json_key(self, first, key, query_env=None):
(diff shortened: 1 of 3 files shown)


mdipierro:
Can you provide some examples of usage (input and output) in the docstring?

leonelcamara:
@mdipierro sure, good point, I'm guessing most people won't want to look at the tests to learn how to use it eheheh. Note that I did accidentally follow the DAL's bad standard of barely having docstrings so we should probably put docstrings on the other functions as well at least those that are in Field, Table

leonelcamara:
@mdipierro done!

PR #563, opened 2019-03-08
Use UTF8 encoding when importing a csv stream

rayluo:
This is one of the approach mentioned in https://github.com/web2py/web2py/issues/2015#issuecomment-470430842

Currently I tested it manually in my Python 3.7 environment.

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -11,6 +11,7 @@ import shutil
 import sys
 import types
 import re
+import io
 from collections import OrderedDict
 from ._compat import (
     PY2, StringIO, BytesIO, pjoin, exists, hashlib_md5, basestring, iteritems,
@@ -881,6 +882,7 @@ class Table(Serializable, BasicStorage):
                              id_offset = None,  # id_offset used only when id_map is None
                              transform = None,
                              validate=False,
+                             encoding="utf-8",
                              **kwargs
                              ):
         """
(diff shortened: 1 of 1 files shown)


mdipierro:
I think this only works in py3. Try use _compat.py

rayluo:
@mdipierro Thanks for your quick response! Would you please be more specific about what did you mean by "use _compat.py"?

* I looked into that file and did not seem to find anything existing that I can use
* Although I did not previously test my current PR on Python 2, I did check [Python 2 documentation and find it also has the TextIOWrapper facility](https://docs.python.org/2/library/io.html#io.TextIOWrapper).
* Lastly, I attempted to install Python 2 into my environment just now, but then my app even failed to start, because there is another exception of `ValueError: unsupported pickle protocol 3` when the app trying to load my model files. This can be a separated topic anyway.

mdipierro:
You are correct that the problem you are having with python 2.7 is not related to your change. It is due to the fact that you session files were created with python 3 and cannot be read by python 2. You have to clear tests. Anyway, travis.ci tests are failing both on py2 and py3 for different reasons. https://travis-ci.org/web2py/pydal/builds/503458861?utm_source=github_status&utm_medium=notification See if you can fix this, else I will fix it later in the week. Thanks for your help.

rayluo:
That travis ci failure seems to be caused by a completely different issue, which presumably exists for more than 2 years (because the latest PR with a green tick was in Jan 2017). I probably would need a long warm-up time to catch up with that history...

mdipierro:
No problem. I can fix it tomorrow

mdipierro:
I do not think this solution works in Python2 as a stream in general is not an io compatible object.
I think the solution proposed by Alex here https://stackoverflow.com/questions/904041/reading-a-utf8-csv-file-with-python is a better one. What to try it?

mdipierro:
I found an alternative solution based on you proposal and this https://stackoverflow.com/questions/904041/reading-a-utf8-csv-file-with-python
It is in this commit:
ebe2e7d221831c9646cd21f552b12f788839890
Please help me check it. Thanks for your help.

TACO-Chloe:
python3.6
web2py 2.18.5

This problem seems to still exist.

Issue #548, opened 2019-01-07
`Cascade` of foreign keys still doesn't work


valq7711:
There is old bug:
```python
db.define_table('a_tbl', Field('name'))
db.define_table('b_tbl', Field('id', 'reference a_tbl'), Field('name'), primarykey=['id'])
# next line causes an error, because b_tbl_id references to b_tbl.id that is reference too
db.define_table('c_tbl', Field('b_tbl_id', 'reference b_tbl'), Field('name'))  
```
The cause of the bug is the following logic
https://github.com/web2py/pydal/blob/5c3355583ce96777539aa13a59842671f2cf7f4c/pydal/migrator.py#L79-L83

that assumes `rfield.type` can not be 'reference ...'

As far as I understand,  it should walk up on the reference fields until the underlying db-type is encountered, no?

PR #578, opened 2019-06-18
Allow reference to the field that is reference too

valq7711:
fix issue #548
This patch is not perfect and has a limitation: referenced field must be `id` and  `integer`.
Solution for other field types requires storing underlying (real) type in the field object.


valq7711:
Is something wrong with my PR?

PR #584, opened 2019-07-18
Add delete_naive

dlage:
Prevent hooks from being executed while (before and after delete) to make it easier to delete from one hook, for example.

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -2423,6 +2423,15 @@ class Set(Serializable):
         ret and [f(self) for f in table._after_delete]
         return ret
 
+    def delete_naive(self):
+        """
+        Same as delete but does not call table._before_delete and _after_delete
+        """
+        db = self.db
+        table = db._adapter.get_table(self.query)
+        ret = db._adapter.delete(table, self.query)
+        return ret
+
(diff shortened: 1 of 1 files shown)


mdipierro:
I agree we need a syntax to do this. I am not convinced this is the best one. It treats delete as a special case. insert, update have the same issue. Can we do naive_insert, naive_update, and naive_delete instead? Also, shouldn't delete call naive_delete?

dlage:
> I agree we need a syntax to do this. I am not convinced this is the best one. It treats delete as a special case. insert, update have the same issue. Can we do naive_insert, naive_update, and naive_delete instead? Also, shouldn't delete call naive_delete?

There already is an update_naive, so I followed the same pattern.

Calling the xxx_naive method from the other one would redo the code block of getting the table. I opted for not doing that both for not increasing complexity of the methods and keeping the same approach as the update_naive.

Regarding the insert_naive I'm not sure there is a need for that but it might make sense in the following case:
* Insert A in table X
* When inserting A insert 1/A in the same table (using insert_naive)
* Stop there

Hadn't thought of a use for it before but yeah, it might come in handy.

Issue #392, opened 2016-07-12
TypeError: unhashable type: 'Field'


ilvalle:
Running python 3.5 I got the following
```
>>> from pydal.objects import Field
>>> f=Field('test')
>>> f in {}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'Field'
```
This snipped code is an example of what there is in web2py https://github.com/web2py/web2py/blob/master/gluon/sqlhtml.py#L80

gi0baro:
I think this is related to changes to the built-in `object.__hash__` method in python 3 compared to python 2, as from https://docs.python.org/3/reference/datamodel.html#object.__hash__:

> A class that overrides `__eq__()` and does not define `__hash__()` will have its `__hash__()` implicitly set to `None`. When the `__hash__()` method of a class is `None`, instances of the class will raise an appropriate `TypeError` when a program attempts to retrieve their hash value, and will also be correctly identified as unhashable when checking `isinstance(obj, collections.Hashable)`.

`Field` inherits from `Expression` that does override `__eq__`.
I think a solution could be add to `Field` (but needs tests):

```python
def __hash__(self):
    return hash(str(self))
```

gi0baro:
Additionally, a test regarding `__hash__` on current pyDAL, with python 2.7:

```python
>>> Post.title.__hash__
<method-wrapper '__hash__' of Field object at 0x1043a1f10>
>>> Post.title.__hash__()
272867825
```

and python 3.5
```python
>>> Post.title.__hash__
>>> Post.title.__hash__()
Traceback (most recent call last):
  File "<console>", line 1, in <module>
TypeError: 'NoneType' object is not callable
```

ilvalle:
Initially, I thought to fix it adding
```    
def __hash__(self):
    return id(self)
```
However, for me is also ok in hashing the string representation.

mdipierro:
I prefer the id(...) than hash(str(...))

PR #587, opened 2019-08-13
Fix TypeError: unhashable type: 'Row'

timnyborg:
In python3, Row's lack of a __hash__ function prevents you from using Rows as a subquery:
`db(db.field.belongs(rows_object))` results in `TypeError: unhashable type: 'Row'`

Similar issue to #392


mdipierro:
I am going to do this in a different way

timnyborg:
Thanks

PR #590, opened 2019-08-31
IS_IN_DB(..., multiple=(..., upper)) becomes inclusive

rayluo:
According to [docs](http://web2py.com/books/default/chapter/29/07/forms-and-validators#IS_IN_DB), the original design intention should be an inclusive lower and upper bound.

Please comment on whether we want to change the documentation or change the implementation. If we agree to change the implementation here, I'll go ahead to adjust the test cases too.

### DIFF
--- a/pydal/validators.py
+++ b/pydal/validators.py
@@ -651,7 +651,7 @@ class IS_IN_DB(Validator):
                 values = new_values
 
             if isinstance(self.multiple, (tuple, list)) and \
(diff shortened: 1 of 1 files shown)


mdipierro:
Sorry this is a breaking of backward compatibility. It is custom to exclude the upper value in ranges in python. It would be better to fix the documentation than change behavior.

PR #615, opened 2020-03-16
Update db2.py

mau86:
Correct lastrowid function, taking in account existence of table._rname
Correct encoding issues setting 'utf-8' encoding in db2:pyodbc connector
Use correct SQL syntax (at least for DB2 for i) in test_connection function


mau86:
Checks fail because none of them uses db2 adapter

PR #624, opened 2020-04-23
Fixed aspects of pydal for preliminary "google:datastore" 

tolgakonik:
- patched _gae to support google.cloud instead of  google.appengine.ext, which is not supported for python 3.
- updated google adapter, to ignore unsupported optional arguments, in cases they are not used. 
- one minor python3 migration fix long -> int

I tested running datastore emulator locally:
gcloud beta emulators datastore start --data-dir=localdata

Using these for access:

os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = 'some-path/some-key-file.json'
os.environ['DATASTORE_EMULATOR_HOST'] = "localhost:8081"

One new thing is that this library requires a new way of defining permission context. So I used the following for the wsgi app (there may be a better way to integrate with py4web):

def ndb_wsgi_middleware(wsgi_app):
    def wrapper(environ, start_response):
        with client.context():
            return wsgi_app(environ, start_response)

    return wrapper

app = ndb_wsgi_middleware(py4web.core.bottle.default_app())


mdipierro:
@lucadealfaro You have looked into this more than I have. thoughts about this PR.

mdipierro:
Was this tested in both py2 and py3? I am afraid it will break py2

lucadealfaro:
So, this uses for Python3 the port to ndb to Python3, which on the one hand is almost compatible with ndb for Python2, on the other hand, Google is already discouraging its use. 

I am writing nqdb, an almost clone of ndb based on the "datastore" interface.  I will open source it soon, just need to test the queries.  The datastore interface is kind of prehistoric, but it is what Google advises for new applications going forward, and nqdb would work likely both for Python 2 and 3, so that could be an alternative way of solving the problem. 

So I have no strong opinions on the best course forward.

jparga:
I don't know if it is more difficult, but would it be better to make Firestore the python3 reference Google Appengine database to pydal instead of Datastore ? 
Could it be treated as a new approach to python3 and let current google:datastore+nbd to Python2 as it is in pydal?

lucadealfaro:
Google advises using Firestore in Datastore mode for server-based data storage: https://cloud.google.com/appengine/docs/standard/python/using-cloud-datastore

So given this, there are (at least) three ways of talking to Datastore in Python 3 Appengine: 

- Using the google.cloud.datastore Python API: https://googleapis.dev/python/datastore/latest/client.html .  This is a somewhat low-level interface, which has been stable for a long time, but is somewhat archaic in syntax (records are represented as dictionaries without any structure, etc).  It does not allow queries containing ORs. 
- Using the new ndb interface for Python 3: https://googleapis.dev/python/python-ndb/latest/index.html .  This is an almost drop-in replacement for ndb that works with Python 3. I think that this is what @tolgakonik is doing.  
- I am writing a simple port of ndb on top of datastore, https://github.com/lucadealfaro/nqdb . Being written on top of datastore, it does not allow for queries containing ORs (at the moment); for that I would need to write a portion using protocol buffers as the interface. 

I think the latter two approaches are the best.  Using the new ndb as done by @tolgakonik is very appealing, because it allows for ORs and ANDs in queries.  The one uncertainty is that Google recommends datastore for new projects rather than ndb, not sure why (long term support?). 

My own nqdb library is very small and simple to maintain, so we can maintain it in open source even if Google stops supporting the new ndb, but being built on datastore, for the moment, there is no simple way to add support for OR queries.  In other words, the GQL of ndb is not supported. (I built nqdb for another purpose, not really for pydal, btw).

jparga:
Hi,
I have received this warning in gae deployed aplications:

_Please remove any imports of google.appengine.api.rdbms.  First Generation Cloud SQL instances have been shut down, and rdbms.py will be removed in a future release.  See: https://cloud.google.com/sql/docs/mysql/deprecation-notice (/base/alloc/tmpfs/dynamic_runtimes/python27g/534e28e62871e9ae/python27/python27_lib/versions/1/google/appengine/api/rdbms.py:33)_

The libray google.appengine.api.rdbms is imported at least, in pydal/_gae.py

`# -*- coding: utf-8 -*-

try:
    from new import classobj
    from google.appengine.ext import db as gae
    from google.appengine.ext import ndb
    from google.appengine.api import namespace_manager, rdbms
    from google.appengine.api.datastore_types import Key  # for belongs on ID
    from google.appengine.ext.ndb.polymodel import PolyModel as NDBPolyModel
except ImportError:
    gae = None
    Key = None`

jparga:
I have tried https://github.com/lucadealfaro/nqdb  and I liked very much its simplicity and performance. I have no answer for Luca doubt about best course forward.
I have realised that for the more simple web services that requires a NoSQL database all the trhee solutions proposed are good enough, even without pydal integration. That would be appropriate for the projects I am working on. 

So to avoid pydal error I tried with google:MySQLdb that works with no problem and, at the same time use Luca nqdb proposal for the specific entities that take advange of a NoSQL perfomance.

It can be messy for medium or big projects, but I wonder if it is usefull to split the data that has to be widely available to be read and shown, from the data that is useful to make control and analisys work. Just thinking about it, because the services option in Google App Engine seems to be suitable to use it.

lucadealfaro:
I wrote nqdb for some other purpose (unrelated to pydal), and I think in order to extend pydal's compatibility to Python3 and Google Datastore, it is better to use the new ndb port by Google, than my nqdb.  The reason, as I perhaps stated before, is that ndb deals with both AND and OR queries, whereas nqdb, being built on the datastore Python API, only provides AND. 

In general, Google is moving away from Datastore.  Even its current offering is a layer on top of Spanner.  But having a compatibility layer for Datastore may still be desirable.  I agree with tolgakinok that SQL on google cloud with pydal works wonderfully. 

So I leave to Massimo and tolgakonik obviously to decide what to do, but after reflection, my impression is that if we wish to use Datastore from Pydal in Python3, then the tolgakonik path via ndb is the best option.

jparga:
Hello, is the Google/NoSQL and Google/NoSQL/NDB adaptation still in progress ?

PR #628, opened 2020-05-14
issue/611 firebird adapter test

villas:


villas:
Proof:  http://www.firebirdfaq.org/faq30/

PR #642, opened 2020-10-06
Enabled ignore_common_filters on IS_IN_DB

agrogers:
IS_NOT_IN_DB('...',  ignore_common_filters=True) had already been implemented but IS_IN_DB had not.

Example:
db.auth_user.SelectedOrganisationID.requires = IS_IN_DB(db, 'Organisation.id', '%(LongName)s', ignore_common_filters=True)

Actually, I really wanted to use the Set method (db(query, ignore_common_filters=True)) in IS_IN_DB but it does not seem to work.

### DIFF
--- a/pydal/validators.py
+++ b/pydal/validators.py
@@ -565,6 +565,7 @@ class IS_IN_DB(Validator):
         left=None,
         delimiter=None,
         auto_add=False,
+        ignore_common_filters=False,
     ):
 
         if hasattr(dbset, "define_table"):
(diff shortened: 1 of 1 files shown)


leonelcamara:
This should not be merged because it does this while building the set: ```table._common_filter = None``` and doesn't set the filters back afterwards. Meaning the table will lose its filters after being used by IS_IN_DB causing potential security problems depending on how the filters are being used by an application.

mdipierro:
Based on leonel's comment above I will decline this

PR #644, opened 2020-11-12
Fix indbmigrator problems

leonelcamara:
Fixes a problem with things being quoted twice due to them being quoted in the query then by the adapters in executesql where the placeholders also get quoted causing an error like this one:

    SyntaxError: syntax error at or near "applications"
    LINE 1: ...RT INTO web2py_filesystem(path,content) VALUES (''applicatio...

Fixes a py3 problem where you would have

        i = self.data.find("\n", self.p) + 1
    TypeError: argument should be integer or bytes-like object, not 'str'

### DIFF
--- a/pydal/helpers/classes.py
+++ b/pydal/helpers/classes.py
@@ -14,9 +14,10 @@ from .._compat import (
     iterkeys,
     itervalues,
     iteritems,
+    to_bytes,
     long,
 )
-from .._compat import to_bytes
+
 from .._globals import THREAD_LOCAL
 from .serializers import serializers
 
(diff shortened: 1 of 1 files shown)


pydal/helpers/classes.py
```diff
@@ -552,7 +553,7 @@ def close_connection(self):
             self.db.executesql(
                 "DELETE FROM web2py_filesystem WHERE path='%s'" % self.filename
             )
-            query = "INSERT INTO web2py_filesystem(path,content) VALUES ('%s','%s')"
+            query = "INSERT INTO web2py_filesystem(path,content) VALUES (%s, %s)"
```
dlage, line 556: I think this fix should be:
```
query = "INSERT INTO web2py_filesystem(path,content) VALUES ('?', '?')"
```

mdipierro: I will take this PR but you are right we should replace all '%s' with '?'. It also requires some refactoring. I will take another PR with that. :-)

PR #645, opened 2020-11-19
Revert "Update upload file name encoding"

mdipierro:
Reverts web2py/pydal#640 because breaks tests (adds / in filenames)

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -29,6 +29,7 @@ from ._compat import (
     copyreg,
     reduce,
     to_bytes,
+    to_native,
     to_unicode,
     long,
     text_type,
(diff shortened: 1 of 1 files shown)


jvanbraekel:
@mdipierro  All test are passing on my side (40 out of 41 because one is skipped).  I don't have any / added neither when I use it on my applications.  I am using Ubuntu 16.04 . Could it be specific to Windows ? Sould I do a new PR ?

mdipierro:
The problem is that b64encode can generate "/" in the filename. It occurs on travis tests. That can beak things. I would take a different PR if you can solve the problem but I think you cannot use b64encode/decode.

jvanbraekel:
Can you provide me a problematic example, to allow me to test it without Travis running locally ?  Without being able to reproduce the issue, i struggle to fix it.

niphlod:
there's a urlsafe_b64encode exactly for this kind of problems. 
"""
Encode bytes-like object s using the URL- and filesystem-safe alphabet, which substitutes - instead of + and _ instead of / in the standard Base64 alphabet, and return the encoded bytes. The result can still contain =
"""

just 2 cents for windows, 
image

jvanbraekel:
Good morning,

Thank you for your help. 'url_safe_b64encode' is the reason why I was asking for a specific test, to allow me to confirm that using it will solve the issue, before doing the PR.  I assume that  windows file name restrictions, can indeed be a hint on why some test may be failing, good catch ! I will update the code and do a new PR.

Issue #671, opened 2021-10-04
Fix unable to parse csv file in web2py break  CSV parsing when dynamic upload is used


jvanbraekel:
Hello,

The PR #654, actually break the CSV parsing,  when it was working previously... Specifically, in case of dynamic upload in web2py  : when the CSV file is uploaded trough the application before import into the database.  TextIOWrapper is then already used to convert the binary stream into a string one the second usage in csv.reader raise an exception. 

An example of usage in web2py controller typically looks  like this :   db.import_from_csv_file(TextIOWrapper(request.vars.file_selector.file,encoding='utf8'))

Removing TextIOWrapper from the call don't work,  because pydal, 'object' and 'base' class are expecting string  and not binary data. Consequently, all the parsing/format tests check for hard-coded string values ( for instance "TABLE' and not b"TABLE"), thus crash when binary data is used.

PR #672, opened 2021-10-04
Only use TextIOWrapper if needed

jvanbraekel:
Fix the exception when TextIOWrapper  is used twice.  Happens if IO stream have to be converted before, for instance when CSV file are uploaded dynamically. See issue #671 for more details.


jvanbraekel:
Duplicate, I have isolate the change in a new branch for the PR.

PR #673, opened 2021-10-05
Only use TextIOWrapper if needed

jvanbraekel:
Fix the exception when TextIOWrapper is used twice. Happens if IO stream have to be converted before, for instance when CSV file are uploaded dynamically. See issue #671 for more details.

### DIFF
--- a/pydal/objects.py
+++ b/pydal/objects.py
@@ -254,7 +254,7 @@ class Row(BasicStorage):
 
 
 def pickle_row(s):
-     return Row, ({k: v for k, v in s.items() if not callable(v)},)
+    return Row, ({k: v for k, v in s.items() if not callable(v)},)
 
 
 copyreg.pickle(Row, pickle_row)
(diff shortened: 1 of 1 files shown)


pydal/objects.py
```diff
@@ -100,7 +100,7 @@
 
 def csv_reader(utf8_data, dialect=csv.excel, encoding="utf-8", **kwargs):
     """like csv.reader but allows to specify an encoding, defaults to utf-8"""
-    csv_reader = csv.reader(TextIOWrapper(utf8_data,encoding), dialect=dialect, **kwargs)
+    csv_reader = csv.reader(utf8_data if isinstance(utf8_data,TextIOWrapper) else TextIOWrapper(utf8_data,encoding), dialect=dialect, **kwargs)
```
mdipierro, line 103: this is a very long line. May I ask you to break it in few lines? Maybe use block for reformatting the file.

jvanbraekel: Done.

PR #688, opened 2022-06-20
Fix IS_IN_DB with multiple=True for single value

robinvandernoord:
When using a py4web grid with Select Widget and choosing only one option while multiple = True, the form gives a validation error.
This is because in this case, the `value` passed to validate is an integer, while `theset` is a list of strings. By explicitly converting `v` to a string, saving the multiple select works again. I hope this does not break any other logic in the module.


mdipierro:
Good catch!

Issue #314, opened 2015-12-01
Syntax error with SQLite3 and PyDAL


MaxMorais:
Hi All!

I have a very silly model, and I'm getting the below traceback

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/base.py", line 174, in __call__
    obj = super(MetaDAL, cls).__call__(*args, **kwargs)
  File "esquema/lds.py", line 70, in __init__
    self.define_schema("Schema", "Entity")
  File "esquema/lds.py", line 93, in define_schema
    obj = cls(self)
  File "esquema/lds.py", line 219, in __init__
    self.define_table()
  File "esquema/lds.py", line 238, in define_table
    primarykey=["name"]
  File "/home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/base.py", line 834, in define_table
    table = self.lazy_define_table(tablename,*fields,**args)
  File "/home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/base.py", line 873, in lazy_define_table
    polymodel=polymodel)
  File "/home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/adapters/base.py", line 482, in create_table
    self.create_sequence_and_triggers(query, table)
  File "/home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/adapters/base.py", line 1366, in create_sequence_and_triggers
    self.execute(query)
  File "/home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/adapters/base.py", line 1385, in execute
    return self.log_execute(*a, **b)
  File "/home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/adapters/base.py", line 1379, in log_execute
    ret = self.get_cursor().execute(command, *a[1:], **b)
sqlite3.OperationalError: near "parent_name": syntax error
>>> pm()
> /home/maxwell/Development/github.com/maxmorais/esquema-env/lib/python2.7/site-packages/pydal/adapters/base.py(1379)log_execute()
-> ret = self.get_cursor().execute(command, *a[1:], **b)
(Pdb) query
*** NameError: name 'query' is not defined
(Pdb) up
[...]

PR #691, opened 2022-07-13
Fix: Syntax error with SQLite3 and PyDAL #314

XpycTee:
Fixed the syntax error that SQLite throws when the foreign key declaration was inserted immediately after the column declaration.
Issue: #314 

I moved the declaration of foreign keys to the end of the command.

I'm not sure it won't break anything, but now I have no problems with this in my project.

### DIFF
--- a/pydal/migrator.py
+++ b/pydal/migrator.py
@@ -29,6 +29,7 @@ class Migrator(object):
         db = table._db
         table._migrate = migrate
         fields = []
+        reffs = str()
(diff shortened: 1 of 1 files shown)


mdipierro:
Did you try both python2 and python3?

XpycTee:
> Did you try both python2 and python3?

No, I tried it only on Python3, it seems to me that it should work in Python2, because the change is the same for both versions

PR #697, opened 2022-08-10
Web2py filesystem issues

### LINKED ISSUE #696: web2py_filesystem Updates/Corrections
1. Attempting to store DAL table structure details in the database rather than the filesystem is currently failing due to an encoding typo where the filename is being inadvertently stored as bytes rather than it's filename in one section, which results in that file not being found when it's queried for and thereby attempts to recreate tables that exist and fails as it's violating the web2py_filesystem primary keys.

2. Binary types in PostgreSQL are "BYTEA" but the web2py_filesystem table creation script has it hardcoded to "BLOB" for all three applicable database types ("mysql", "postgres", "sqlite",) which causes the web2py_filesystem table creation to fail when starting py4web.

macneiln:
Logic to resolve #696.

### DIFF
--- a/pydal/helpers/classes.py
+++ b/pydal/helpers/classes.py
@@ -488,7 +488,7 @@ class DatabaseStoredFile:
                 raise NotImplementedError(
                     "DatabaseStoredFile only supported by mysql, potresql, sqlite"
                 )
-            blobType = "BLOB" if not db._adapter.dbengine == "postgres" else "BYTEA"
+            blobType = "BYTEA" if db._adapter.dbengine == "postgres" else "BLOB" 
             sql = "CREATE TABLE IF NOT EXISTS web2py_filesystem (path VARCHAR(255), content %(blobType)s, PRIMARY KEY(path));" % {"blobType": blobType}
             if db._adapter.dbengine == "mysql":
                 sql = sql[:-1] + " ENGINE=InnoDB;"
--- a/pydal/helpers/classes.py
+++ b/pydal/helpers/classes.py
@@ -488,7 +488,8 @@ class DatabaseStoredFile:
                 raise NotImplementedError(
                     "DatabaseStoredFile only supported by mysql, potresql, sqlite"
                 )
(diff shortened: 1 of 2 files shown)


pydal/helpers/classes.py
```diff
@@ -488,7 +488,8 @@ def try_create_web2py_filesystem(db):
                 raise NotImplementedError(
                     "DatabaseStoredFile only supported by mysql, potresql, sqlite"
                 )
-            sql = "CREATE TABLE IF NOT EXISTS web2py_filesystem (path VARCHAR(255), content BLOB, PRIMARY KEY(path));"
+            blobType = "BYTEA" if db._adapter.dbengine == "postgres" else "BLOB" 
+            sql = "CREATE TABLE IF NOT EXISTS web2py_filesystem (path VARCHAR(255), content %(blobType)s, PRIMARY KEY(path));" % {"blobType": blobType}
```
mdipierro, line 492: This change is problematic. It will may break apps of current users if the table already exists. Can you explain why it is necessary?

pydal/migrator.py
```diff
@@ -315,7 +315,11 @@ def create_table(self, table, migrate=True, fake_migrate=False, polymodel=None):
             query = "CREATE TABLE %s(\n    %s\n)%s" % (table_rname, fields, other)
 
         uri = self.adapter.uri
-        if uri.startswith("sqlite:///") or uri.startswith("spatialite:///"):
+        if isinstance(self, InDBMigrator):
```
mdipierro, line 318: good catch!

pydal/helpers/classes.py
macneiln: The "BLOB" field type doesn't exist in PostgreSQL, it's type "BYTEA", so currently it just errors attempting to create the web2py_filesystem table with "BLOB type does not exist" since it's not available in PostgreSQL. 
[https://www.postgresql.org/docs/current/datatype-binary.html](https://www.postgresql.org/docs/current/datatype-binary.html)

PR #729, opened 2025-02-04
Massively simplify join logic

laundmo:
I stumbled across this while investigating why the dal generated a cross join when i wanted a left join.

This PR adds a check to this, raising a ValueError if a joins ON condition doesn't contain the table name at all (very naive, but should catch the mistake in almost all cases)

It also vastly simplifies the join code, by collecting all the required info beforehand and deduplicating the join building.

### DIFF
--- a/pydal/adapters/base.py
+++ b/pydal/adapters/base.py
@@ -626,35 +626,51 @@ class SQLAdapter(BaseAdapter):
             field = field.st_astext()
         return self.expand(field, query_env=query_env)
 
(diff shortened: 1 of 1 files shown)


laundmo:
Ah, of course my naive check for "hey, use an alias here!" fails for self-referencing tables... but i think i can check for that? sure.

laundmo:
hi @mdipierro, i noticed you reverted this PR again recently. Could you share your reasoning?

PR #731, opened 2025-04-10
Include already existing JSON operators in other postgres adapters and fix contains for bigint[]

jgrazina:
- Multiple postgres adapters were supporting JSON/JSONB columns, but were missing out on JSON operators defined in PostgreDialectJSON dialect
- Contains for bigint[] was using ilike incorrecly


mdipierro:
thanks!

Issue #746, opened 2025-10-24
Lazy tables broken by the new get_default_validator


leonelcamara:
Specifically here

https://github.com/web2py/pydal/blob/8bed9b7537b4b3f843ecc57d960a719554741d97/pydal/objects.py#L113C1-L120C1

The need to find out the id field of the referenced table forces you to define the table which causes migration problems and problems with self references.

PR #750, opened 2025-11-15
Revert "Fix: #746"

mdipierro:
Reverts web2py/pydal#749

We already fixed it in 

```
if not self._db or tablename not in self._db:
             # The table being referenced is not defined yet
             return None
```


leonelcamara:
Nope, a tablename can be "in" db as in __contains__ says it's there and still throw an AttributeError when you try to access it

mdipierro:
I will unrevert the revert.

PR #771, opened 2026-06-02
Fixes a bug in postgrtes array representers

leonelcamara:


mdipierro:
was this a regression because of recent changes or a pre-existing condition?

leonelcamara:
> was this a regression because of recent changes or a pre-existing condition?

@mdipierro It's pre-existing

